PR #2, opened 2018-07-25
Fix cluster solver check

gbalduzz:
The covariance matrix should be computed against the sample average, not the expected value, as it was done in the statistical validation test. This PR fix this issue in the cluster_solver_check application.

### DIFF
--- a/test/unit/parallel/mpi_concurrency/mpi_collective_sum_test.cpp
+++ b/test/unit/parallel/mpi_concurrency/mpi_collective_sum_test.cpp
@@ -205,6 +205,7 @@ TEST_F(MPICollectiveSumTest, ComputeCovarianceScalar) {
   using CovarianceDomain = dca::func::dmn_variadic<FunctionDomain, FunctionDomain>;
 
   dca::func::function<double, CovarianceDomain> covariance_test("test");
+  dca::func::function<double, CovarianceDomain> covariance_test2("test2");
   dca::func::function<double, CovarianceDomain> covariance_expected("expected");
 
   dca::func::function<double, FunctionDomain> f("f");
@@ -219,6 +220,10 @@ TEST_F(MPICollectiveSumTest, ComputeCovarianceScalar) {
 
   sum_interface_.computeCovariance(f, f_mean, covariance_test);
 
+  sum_interface_.computeCovarianceAndAvg(f, covariance_test2);
+  for (int i = 0; i < f_mean.size(); ++i)
+    EXPECT_DOUBLE_EQ(f_mean(i), f(i));
+
   covariance_expected(0, 0) = covariance_expected(0, 1) = covariance_expected(0, 2) =
       covariance_expected(0, 3) = 0;
   covariance_expected(1, 1) = 5.25;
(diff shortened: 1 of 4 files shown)


gbalduzz:
Can this be merged?

test/unit/parallel/mpi_concurrency/mpi_collective_sum_test.cpp
```diff
@@ -205,6 +205,7 @@ TEST_F(MPICollectiveSumTest, ComputeCovarianceScalar) {
   using CovarianceDomain = dca::func::dmn_variadic<FunctionDomain, FunctionDomain>;
 
   dca::func::function<double, CovarianceDomain> covariance_test("test");
+  dca::func::function<double, CovarianceDomain> covariance_test2("test2");
```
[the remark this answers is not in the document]
gbalduzz: Changed the variable name and added a description. 0341b2e21bfb2bb5765925b7b25872bb8395e66f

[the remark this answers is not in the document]
yingwaili: I have a suggestion: how about renaming covariance_test2 (or covariance2) to covarianceAndAvg_test (or covarianceAndAvg) to better reflect what the test is doing?

include/dca/phys/dca_step/cluster_mapping/coarsegraining/coarsegraining_sp.hpp
```diff
@@ -143,7 +143,7 @@ class coarsegraining_sp : public coarsegraining_routines<parameters_type, K_dmn>
 
 private:
   parameters_type& parameters;
-  concurrency_type& concurrency;
```
uhaehner: I think this slipped into the PR.

test/unit/parallel/mpi_concurrency/mpi_collective_sum_test.cpp
```diff
@@ -270,8 +290,10 @@ TEST_F(MPICollectiveSumTest, ComputeCovarianceComplex) {
     for (int j = i + 1; j < 2 * f.size(); j++)
       covariance_expected(j, i) = covariance_expected(i, j);
 
-  for (int i = 0; i < covariance_test.size(); i++)
-    EXPECT_DOUBLE_EQ(covariance_expected(i), covariance_test(i));
+  for (int i = 0; i < covariance.size(); i++) {
```
uhaehner: Maybe explain why the results are the same.

gbalduzz: I don't think an explanation for why a method called compute A, and one called compute A and B, provide the same A.

PDoakORNL: So do we still need to maintain computeCovariance? Do we ever call with a different estimate than the average?

gbalduzz: No, not really at the moment. It was written by Andrei in case we need to compute (x-y)(x-y)^t with a generic y != <x>. But currently, there is no test that needs it.

PDoakORNL:
So I thought this whole thing had to do with, but it doesn't appear to.
We in fact seem to avoid applying the 1/(N-1) or 1/(N) prefactor.

gbalduzz:
We apply the 1/N factor with the method sum_and_average. We need the ML estimator, not the ubiased one (with 1/N-1).

PDoakORNL:
Ok I see, got thrown off by the multiple calls to sum_and_average. Ok I'm fine with this being merged.

uhaehner:
retest this please

uhaehner:
retest this please

uhaehner:
Ready to merge if tests pass.

PR #6, opened 2018-08-19
Hdf5 write const vector

gbalduzz:
Minor bug in HDF5 writer. Without const correct arguments it is not possible to write a const container.


include/dca/io/hdf5/hdf5_writer.hpp
```diff
@@ -54,58 +54,60 @@ class HDF5Writer {
...
   template <typename arbitrary_struct_t>
-  static void to_file(arbitrary_struct_t& arbitrary_struct, std::string file_name);
+  static void to_file(const arbitrary_struct_t& arbitrary_struct, std::string file_name);
 
   template <typename scalar_type>
   void execute(std::string name, scalar_type value);
 
   template <typename scalar_type>
-  void execute(std::string name, std::pair<scalar_type, scalar_type>& value);
+  void execute(std::string name, const std::pair<scalar_type, scalar_type>& value);
```
biddisco: For correctness (and efficiency), I suspect that all those `std::string name` vars in the arguments should be `const std::string &name`. Might be a good idea to update them as well

gbalduzz: const ref and deep copy are functionally equivalent here, and given that is by far not performance critically, it is the same. But sure, I can mod it.

uhaehner:
retest this please

uhaehner:
retest this please

uhaehner:
How about renaming the file hdf5_reader_test.cpp as it now tests both hdf5 reader and writer?

gbalduzz:
Merged the reader and writer tests in e6a74c8637e06b48aa146793e8764087ee999393.

uhaehner:
retest this please

uhaehner:
Ready to be merged if tests pass.

uhaehner:
retest this please

PR #8, opened 2018-08-23
Some platforms see failed tests with smaller numbers of measurements

PDoakORNL:
Stand alone, needed for cross platform testing


biddisco:
Even with this patch, I see 
```
The following tests FAILED:
         86 - bilayer_lattice_Nc1_interband (Failed)
         87 - bilayer_lattice_Nc1_intra_plus_interband (Failed)
         88 - bilayer_lattice_Nc1_intraband (Failed)
Errors while running CTest
```
but reducing the tolerance on the tests from 1e-12 to 1e-11 does fix them.

gbalduzz:
Wait, don't we need to merge the parameter change measurements_per_process_and_accumulator_ -> measurements_ before doing this?

gbalduzz:
@PDoakORNL: should I cherry pick the changes to the measurements parameter or have you already done work in this direction?

PDoakORNL:
right this should wait until those go in.  So post accumulator...
I'm going to close this for now.

PR #10, opened 2018-08-24
Link external libs at the end.

gbalduzz:
With this reorder I am able to build on my laptop, as MAGMA requires the linear algebra libraries in the DCA_EXTERNAL_LIBS list.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,7 +154,7 @@ if (DCA_HAVE_CUDA)
     ${DCA_CUDA_LIBS})
 endif()
 
-# The BLAS and LAPACK librearies in DCA_EXTERNAL_LIBS should be linked after MAGMA.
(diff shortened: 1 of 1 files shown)


PDoakORNL:
test this please

CMakeLists.txt
```diff
@@ -141,20 +141,21 @@ set(DCA_LIBS
...
-    cuda_utils) 
+    cuda_utils)
   list(APPEND DCA_LIBS
     ctaux_walker_kernels
     lapack_kernels
     blas_kernels
     ${DCA_CUDA_LIBS})
 endif()
 
+list(APPEND DCA_LIBS ${DCA_EXTERNAL_LIBS})
```
yingwaili: Perhaps add a comment before this line to document the reason for doing this?  Something like:
"# Link external libraries at the end because MAGMA requires the linear algebra libraries in the DCA_EXTERNAL_LIBS list."

Otherwise the reason for doing so will be forgotten overtime.

PR #12, opened 2018-08-24
[WIP] Unstable tests need to be removed from normal testing

PDoakORNL:
For purposes of have clean CI for master and PR's

This PR adds a new category of tests UNSTABLE result, NiO Stats does
not run without -DDCA_WITH_TESTS_UNSTABLE=ON

This is a crude way to handle this issue, feel free to file an issue for improvement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,6 +161,8 @@ option(DCA_WITH_TESTS_FAST "Build DCA++'s fast tests." OFF)
 option(DCA_WITH_TESTS_EXTENSIVE "Build DCA++'s extensive tests." OFF)
 option(DCA_WITH_TESTS_PERFORMANCE "Build DCA++'s performance tests. (Only in Release mode.)" OFF)
 option(DCA_WITH_TESTS_VALIDATION  "Build DCA++'s validation tests." OFF)
+option(DCA_WITH_TESTS_UNSTABLE "Build DCA++'s unstable statistical tests." OFF)
+
 
 set(TEST_RUNNER "" CACHE STRING "Command for executing (MPI) programs.")
 set(MPIEXEC_NUMPROC_FLAG "-n" CACHE STRING "Flag used by TEST_RUNNER to specify the number of processes.")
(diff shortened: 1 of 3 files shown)


uhaehner:
I think @gbalduzz should revisit the test to make sure that it doesn't fail so often.

Since the test is unstable since it's stochastic, I would add it to DCA_WITH_TESTS_VALIDATION if it's a validation test, or add a category DCA_WITH_TESTS_STOCHASTIC. (sounds better to me than unstable ;))

Runtime is not an issue for me because it's part of the extensive tests.

biddisco:
DCA_WITH_TESTS_STOCHASTIC :+1:

biddisco:
Could we also simply modify the testing procedure to run the stochastic tests N times? If the Prob(fail) is small enough, can we be 99.99% certain that running it 5 times or so would be enough?

The main reason for doing the tests is to see if something has been broken, and usually, a break in the code produces something catastrophic in the results (true?) - so relaxing the tolerances on the test should be enough to allow 99.x% pass rate, but still catch fundamental breakages (also true?)

gbalduzz:
Something similar to `DCA_WITH_TESTS_STOCHASTIC` was proposed, and than rejected by @uhaehner . Though I believe it would be best to clump all the statistical tests under a flag and turn it off by default. This would lead to a speedup in automated tests for merging PRs. If a change does not break the solver and you get the same numerical results, there is no way the statistical test should fail. If you tinker with the solver, and change the numerical results, it is the developer responsibility to validate the changes with the stochastic tests, and update the no-change baselines.
I believe we should 
- Clump all stochastic tests under a CMAKE flag.
- Insert a no-change test for the ss-hyb solver.
- significantly decrease the number of measurements in the no-change tests of ctaux: `bilayer_lattice_Nc1...` and `square_lattice_Nc4...`.

If you agree with this changes I can implement them as they would help with the merge process. The NiO test fails a bit too often, an this might be an indication of an issue.

uhaehner:
If we add DCA_WITH_TESTS_STOCHASTIC, do we get rid of DCA_WITH_TESTS_VALIDATION?

Regarding runtime of tests and tests categories, there are two questions for me:
1.  What should be the runtime limit for each category (DCA_WITH_TESTS_FAST, DCA_WITH_TESTS_EXTENSIVE, DCA_WITH_TESTS_???)?
2. Which tests categories are run for a PR, post-push (master), weekly (and nightly)?

gbalduzz:
Yes, we would use only DCA_WITH_TESTS_STOCHASTIC.

PR #15, opened 2018-08-30
Changed  measurements parameter.

gbalduzz:
Implements the change from measurements_per_process_per_accumulator into measurements. This allows doing the same amount of measurements even if some processors are excluded from the communicator.
EDIT:
~I still have to update the `tutorial/` files, but the code changes can be reviewed.~
After the change is finalized, I will need to update the wiki.

### DIFF
--- a/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_accumulator.hpp
+++ b/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_accumulator.hpp
@@ -311,10 +311,7 @@ void CtauxAccumulator<device_t, parameters_type, Data>::finalize() {
     for (int l = 0; l < M_r_w_stddev.size(); l++)
       M_r_w_stddev(l) = std::sqrt(abs(M_r_w_squared(l)) - std::pow(abs(M_r_w(l)), 2));
 
-    double factor =
-        1. / std::sqrt(1 +
-                       concurrency.number_of_processors() * parameters.get_accumulators() *
-                           parameters.get_measurements_per_process_and_accumulator());
+    double factor = 1. / std::sqrt(parameters.get_measurements() - 1);
 
     M_r_w_stddev *= factor;
   }
--- a/include/dca/phys/dca_step/cluster_solver/stdthread_qmci/stdthread_qmci_accumulator.hpp
+++ b/include/dca/phys/dca_step/cluster_solver/stdthread_qmci/stdthread_qmci_accumulator.hpp
@@ -27,13 +27,12 @@ namespace stdthreadqmci {
 
 template <class qmci_accumulator_type>
 class stdthread_qmci_accumulator : protected qmci_accumulator_type {
-  typedef typename qmci_accumulator_type::my_parameters_type parameters_type;
+  using ThisType = stdthread_qmci_accumulator<qmci_accumulator_type>;
+  using Parameters = typename qmci_accumulator_type::my_parameters_type;
   using Data = typename qmci_accumulator_type::DataType;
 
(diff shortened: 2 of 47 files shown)


include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_accumulator.hpp
```diff
@@ -311,10 +311,7 @@ void CtauxAccumulator<device_t, parameters_type, Data>::finalize() {
     for (int l = 0; l < M_r_w_stddev.size(); l++)
       M_r_w_stddev(l) = std::sqrt(abs(M_r_w_squared(l)) - std::pow(abs(M_r_w(l)), 2));
 
-    double factor =
-        1. / std::sqrt(1 +
-                       concurrency.number_of_processors() * parameters.get_accumulators() *
-                           parameters.get_measurements_per_process_and_accumulator());
+    double factor = 1. / std::sqrt(parameters.get_measurements() - 1);
```
gbalduzz: I am pretty sure the n+1 instead of n-1 was a mistake in the code.

include/dca/phys/dca_step/cluster_solver/stdthread_qmci/stdthread_qmci_accumulator.hpp
```diff
@@ -27,13 +27,12 @@ namespace stdthreadqmci {
...
-  typedef typename qmci_accumulator_type::my_parameters_type parameters_type;
+  using ThisType = stdthread_qmci_accumulator<qmci_accumulator_type>;
+  using Parameters = typename qmci_accumulator_type::my_parameters_type;
   using Data = typename qmci_accumulator_type::DataType;
 
-  typedef stdthread_qmci_accumulator<qmci_accumulator_type> this_type;
-
 public:
-  stdthread_qmci_accumulator(parameters_type& parameters_ref, Data& data_ref, int id);
+  stdthread_qmci_accumulator(Parameters& parameters_ref, Data& data_ref, int meas_t_do, int id);
```
gbalduzz: The need to track the measurements here will disappear if we merge https://github.com/CompFUSE/DCA-develop/pull/191.

uhaehner:
test this please

gbalduzz:
I suppose I will wait for the review to update the wiki.

uhaehner:
I think we should guarantee backward compatibility as the paper and the documentation that comes with it still uses `measurements-per-process-and-accumulator`.

biddisco:
I disagree with the backwards compatibility : I had to make a similar change in my HPX branch to account for the fact that everything is asynchronous and there is no need to enforce each accumulator to perform `getWorkload` number of accumulations. If an accumulator is free and a thread is available, it can just go (a cpu accumulator will run slower than a gpu one and if it does less, so be it). It simplified my code considerably to throw away the rigid counting that is in the current code and simply allow N measurements to be made regardless of who/what makes them. Since walkers are not doing extra walk steps when there are spare resources, the number of walker sweeps is always equal to the number of measurements so the measurements per accumulator is essentially redundant and you just need a total measurement count to be enforced.

Backwards compatibility is useful if you have a lot of users whose code will break and you need to keep them happy. For everyone else, you can provide **release notes** that list changes like this so that a user can skim through the list and be aware that some changes have been made to the API. In this case, an AI change would warrant a minor release number update.

uhaehner:
Merging this should be accompanied by a minor release (v1.0.1) to document the API change.

gbalduzz:
Updated wiki and number of measurements in the tc tutorial. From my side, the PR is ready to be merged.

uhaehner:
test this please

PR #16, opened 2018-09-01
Cuda containers

gbalduzz:
Removes the macro definitions `DCA_MAX_THREADS`. This allows to allocate enough resources to run an arbitrary number of walkers in parallel and limits the number of unused CUDA streams.
I could consider redesigning this to use a std::map<std::pair<int,int>, ...> container to store streams and handles, so that we do not have to explicitly resize to the number of threads, or decide a priory a maximum number of streams per thread. If you think It is a good idea I could file an issue, or implement it if I happen to have the time.

there are also two small modifications I pulled together with the same files in the folder: 
- Avoid multiple initializations of Magma.
- Throw a std::bad_alloc if cuda fails to allocate memory.


uhaehner:
test this please

uhaehner:
retest this please

biddisco:
@gbalduzz The task handler currently only serves as a lookup from 'id' to the RNG to use and for some walker/accumulator decisions. I suspoect we'll still need it or something like it for the RNG if results are to be reproducable, but the thread pool stuff can almost certainly be done without having to label threads as walker/ or accumulator etc. I will worry about this when the thread pool stuff is merged.

PDoakORNL:
@biddisco John the threadpool went in #9 can you clarify what you mean?  Do you actually mean when we are done this push and can begin now work based on the gpu_trunk integration?

biddisco:
@PDoakORNL My bad. I didn't realize the thread pool stuff was in master. What I really meant was that there is no urgency in remove the task handler, but I'll look at ways of getting rid of it. It isn't doing any harm, but it is only used at the start of thread initialization (AFAIK) and probably it is redundant.

PR #22, opened 2018-09-04
Dnfft gpu

gbalduzz:
This PR is necessary to implement the single particle accumulation on the GPU. The test helper will later be used for the tp accumulation tests as well.


src/math/nfft/dnfft_1d_kernels.cu
```diff
@@ -113,7 +115,7 @@ __global__ void sumKernel(const ScalarType* in, const int ldi, ScalarType* out,
   const int i = blockIdx.x * blockDim.x + threadIdx.x;
   const int j = blockIdx.y * blockDim.y + threadIdx.y;
 
-  out[i + ldo * j] += in[i + ldo * j];
+  out[i + ldo * j] += in[i + ldi * j];
```
gbalduzz: This sort of mistakes can be avoided by extending the matrix view class to device code, as it is done in the ct_int branch: https://github.com/CompFUSE/DCA-develop/blob/ct_int/include/dca/linalg/matrix_view.hpp.
Though I suppose we can do it after we merged the rest of gpu_trunk.

uhaehner:
test this please

uhaehner:
retest this please

include/dca/math/nfft/dnfft_1d.hpp
```diff
@@ -48,7 +48,9 @@ class Dnfft1D {
   using ThisType = Dnfft1D<ScalarType, WDmn, PDmn, oversampling, mode>;
   using ElementType = ScalarType;
 
+public:
```
yingwaili: What is the reason of putting a second "public:" here?  I don't think it is necessary

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+                     oversampling, CUBIC> {
+public:
+  using BDmn = func::dmn_0<phys::domains::electron_band_domain>;
+  using PDmn = func::dmn_variadic<BDmn, BDmn, RDmn>;
+
+  using ThisType = Dnfft1DGpu<ScalarType, WDmn, RDmn, oversampling>;
+  using ElementType = ScalarType;
+  using BaseClass = Dnfft1D<ScalarType, WDmn, PDmn, oversampling, CUBIC>;
+
+public:
```
yingwaili: Same as above.

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+  Dnfft1DGpu(double beta, cudaStream_t stream, bool accumulate_m_sqr = false);
+
+  // Resets the accumulated quantities. To be called before each DCA iteration.
+  void initialize();
+
+  // Accumulates asynchronously on the device the entries of the M matrix at times and orbitals
+  // described by the provided configuration.
+  // Postcondition: M and config shall not be modified untill 'synchronizeCopy' is called.
```
yingwaili: typo: untill --> until

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Ensures that the arguments of the method 'accumulate' have been copied to the GPU.
+  void synchronizeCopy() {
+    m_copied_event_.block();
+  }
+
+private:
+  void initializeDeviceCoefficients();
+
+private:
```
yingwaili: Again, I don't think a second private is necessary.

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+  initializeDeviceCoefficients();
+  assert(cudaPeekAtLastError() == cudaSuccess);
+  func::dmn_variadic<BDmn, BDmn, RDmn> bbr_dmn;
+  const int n_times = BaseClass::PaddedTimeDmn::dmn_size();
+  accumulation_matrix_.resizeNoCopy(std::make_pair(n_times, bbr_dmn.get_size()));
+  if (accumulate_m_sqr)
+    accumulation_matrix_sqr_.resizeNoCopy(accumulation_matrix_.size());
+}
+
+template <typename ScalarType, typename WDmn, typename RDmn, int oversampling>
```
yingwaili: From the use of this function, I think it is better renamed as "resetAccumulationMatrix()" or alike.

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+  if (accumulate_m_sqr_)
+    accumulation_matrix_sqr_.setToZero(stream_);
+  assert(cudaPeekAtLastError() == cudaSuccess);
+}
+
+template <typename ScalarType, typename WDmn, typename RDmn, int oversampling>
+void Dnfft1DGpu<ScalarType, WDmn, RDmn, oversampling, CUBIC>::initializeDeviceCoefficients() {
+  static bool initialized = false;
+  static std::mutex mutex;
+  if (initialized)
```
yingwaili: It seems that "initialized" is always false at this point.  What is the purpose of this conditional check?

include/dca/math/nfft/dnfft_1d_gpu.hpp
```diff
@@ -0,0 +1,245 @@
...
+    return;
+
+  std::unique_lock<std::mutex> lock(mutex);
+  if (initialized)
```
yingwaili: Same here.

include/dca/math/nfft/kernels_interface.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <cuda.h>
+
+namespace dca {
+namespace math {
+namespace nfft {
+namespace details {
+// dca::math::nfft::details::
+
+struct ConfigElem {
+  int b, r;
```
yingwaili: It is clearer to rename the variables: b --> band, r --> site / position.

include/dca/math/nfft/nfft_helper.cu.hpp
```diff
@@ -0,0 +1,184 @@
...
+  delta_t = get_tau(t_idx) - t - get_window_tau(t_wnd_idx);
+
+  const int oversampling = parameters_int_[3];
+  const int window_sampling = parameters_int_[4];
+  t_idx -= oversampling;
+  const int delta_tau_index = t_wnd_idx - oversampling * window_sampling;
+  const int j = delta_tau_index % window_sampling;
+  const int i = (delta_tau_index - j) / window_sampling;
+
+  constexpr int interpolation_size = 4;  // Hard coded cubic case.
```
yingwaili: This hard coding might be prone to error in the future.  My suggestion is better to make it dependent on the NfftModeNames.

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Prevent multiple initializations. It might be called by multiple threads multiple time. The first initialized check avoids locking and unlocking the mutex on subsequent iterations.
A cleaner implementation can be achieved with std::call_once, but I was not aware of its existence. I might change it later if I have the time.

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: same pattern as before.

include/dca/math/nfft/kernels_interface.hpp
gbalduzz: renamed in 1eb76429faaefa254286bfda76ff0bc2be6ed0f4

include/dca/math/nfft/nfft_helper.cu.hpp
gbalduzz: used a template parameter in aefb57e04087480b2a95aad4a28b1f22cfb3a435 nad made clear that the kernels are implemented only for the cubic case. Note that the DnfftGpu is defined only for the CUBIC case.

include/dca/math/nfft/dnfft_1d.hpp
gbalduzz: Just separating typedefs from methods. I suppose a space can achieve the same. Changed in 547604f176339765f2a44aac02708f7b0ec5dcb7

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Changed in 547604f176339765f2a44aac02708f7b0ec5dcb7

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Fixed in 547604f176339765f2a44aac02708f7b0ec5dcb7

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Changed in 547604f176339765f2a44aac02708f7b0ec5dcb7

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Renamed to resetAccumulation in e7f136b4333ff98dc2b008cbf1f7694c72a4ea3f to keep the same name in the CPU/GPU version. I might as well us a vector or any other container for the accumulated quantities.

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: Implemented in 1caf9406fcc8c5a1e607ae59764749e88572df10.

include/dca/math/nfft/dnfft_1d_gpu.hpp
gbalduzz: imoplemented in 1caf9406fcc8c5a1e607ae59764749e88572df10.

src/math/nfft/dnfft_1d_kernels.cu
```diff
@@ -0,0 +1,168 @@
...
+}
+
+template <typename ScalarType>
+struct HelperSelector {
+  static NfftHelperManager<ScalarType> value;
+};
+template <typename ScalarType>
+NfftHelperManager<ScalarType> HelperSelector<ScalarType>::value;
+
+struct ConfigElem {
```
gbalduzz: I am not sure why I need to redefine `ConfigElem` even though it should be defined and in scope from `kernels_interface.hpp`, but this is the only way it compiles. Any idea on the cause?

gbalduzz: Solved: DCA_HAVE_CUDA was not defined.

gbalduzz:
Replied to review and merged master.

src/math/nfft/dnfft_1d_kernels.cu
biddisco: The *.cu files are compiled with different options/flags if CUDA_PROPAGATE_HOST_FLAGS is not on. the easiest fix is to just #include dca/config/haves_defines so that the right things are picked up.

include/dca/math/nfft/dnfft_1d_gpu.hpp
yingwaili: I see.  The new implementation is clearer in what the code is doing.

PR #27, opened 2018-09-07
Add parameters for threaded solver and accumulator.

gbalduzz:
Quick PR to add a couple of options.

### DIFF
--- a/include/dca/phys/parameters/mci_parameters.hpp
+++ b/include/dca/phys/parameters/mci_parameters.hpp
@@ -19,6 +19,8 @@
 #include <stdexcept>
 #include <string>
 
+#include "dca/phys/error_computation_type.hpp"
+
 namespace dca {
 namespace phys {
 namespace params {
@@ -33,7 +35,9 @@ public:
         measurements_(100),
         walkers_(1),
         accumulators_(1),
-        adjust_self_energy_for_double_counting_(false) {}
+        shared_walk_and_accumulation_thread_(false),
+        adjust_self_energy_for_double_counting_(false),
+        error_computation_type_(0) {}
 
   template <typename Concurrency>
   int getBufferSize(const Concurrency& concurrency) const;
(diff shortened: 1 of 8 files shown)


uhaehner:
test this please

include/dca/phys/parameters/mci_parameters.hpp
```diff
@@ -86,7 +96,9 @@ class MciParameters {
   int measurements_;
   int walkers_;
   int accumulators_;
+  bool shared_walk_and_accumulation_thread_;
   bool adjust_self_energy_for_double_counting_;
+  int error_computation_type_;
```
biddisco: why is this an int and not an ErrorComputationType

include/dca/phys/parameters/mci_parameters.hpp
```diff
@@ -67,9 +71,15 @@ class MciParameters {
...
     return accumulators_;
   }
+  bool shared_walk_and_accumulation_thread() const {
+    return shared_walk_and_accumulation_thread_;
+  }
   bool adjust_self_energy_for_double_counting() const {
     return adjust_self_energy_for_double_counting_;
   }
+  ErrorComputationType get_error_computation_type() const {
+    return ErrorComputationType(error_computation_type_);
```
biddisco: A converion would not be necessary if this was ErrorComputationType

include/dca/phys/parameters/mci_parameters.hpp
```diff
@@ -177,15 +195,25 @@ void MciParameters::readWrite(ReaderOrWriter& reader_or_writer) {
...
     }
+
     try {
       reader_or_writer.execute("measurements", measurements_);
     }
     catch (const std::exception& r_e) {
     }
 
+    // Read error computation type.
+    std::string error_type = toString(static_cast<ErrorComputationType>(error_computation_type_));
```
biddisco: same here. No real need for a type cast

include/dca/phys/parameters/mci_parameters.hpp
gbalduzz: I would need to add specializations to MPITypeMap for each enumeration. Do you think it is a better idea or would there be a way to specialize the class in a way to match all possible enumerations?

biddisco:
OK. I spent a bit of time experimenting with this and the existing implementation is much more straightforward than using a scoped enumerated type in place of an int in the member function. The simple int/enum conversion is less intrusive than having to add an MPI pack overload. Switching to a simple enumerated type with implicit conversion to int suffers from name collision with FourPointType::NONE and cleaning it all up would be a new PR/Issue.

PR #31, opened 2018-09-11
G4 data

gbalduzz:
Extends the domain of G4 to include the frequency and momentum exchange domains.
func::difference and its test have been modified to allow comparing different functions of the same size. Useful for when the domain has additional dimensions of size 1.


gbalduzz:
Merged master into this.

test/integration/cluster_solver/stdthread_qmci/stdthread_ctaux_tp_test.cpp
```diff
@@ -80,18 +80,16 @@ TEST(StdthreadCtauxTest, GreensFunctions) {
   if (!update_baseline) {
     // Compare to baseline results.
     if (concurrency.id() == 0) {
-      auto G_k_w_check = data.G_k_w;
-      using DomainType = typename Data::TpGreensFunction::this_domain_type;
-      dca::func::function<std::complex<double>, DomainType> G4_check(data.get_G4_k_k_w_w().get_name());
-      G_k_w_check.set_name(data.G_k_w.get_name());
+      Data::SpGreensFunction G_k_w_check(data.G_k_w.get_name());
```
biddisco: It is intentional I assume to use Sp and not Tp here?

gbalduzz: Yes, I am testing both G and G4.

PR #32, opened 2018-09-11
Extend two particle test.

gbalduzz:
Depends on #31 and #28.
This PR tests the two particle accumulation over a 3x3 grid, so that r_i - r_j != r_j - r_i in general. It also uses the same configuration generator as the other tests.


gbalduzz:
Replaced by #33.

PR #33, opened 2018-09-12
Tp accumulator

gbalduzz:
~Depends on #31 and #28.~
It replaces #32 as there is no need to review a change to the same test twice.

It uses the cached ndft and space transform 2D objects to define the two particle accumulation both on the CPU and the GPU. Given that the GPU accumulators share memory, extra care has to be spent in ensuring the shared memory is initialized only once per DCA iteration.


uhaehner:
retest this please

uhaehner:
retest this please

biddisco:
@PDoakORNL @gbalduzz Understood. I 'fixed' the CallOncePerLoop without really addressing what it was actually being used for in this particular case. I'll take another look at it the initialization with your comments in mind.

gbalduzz:
.*test\W+this\W+please.*

gbalduzz:
retest\W+this\W+please

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp
```diff
@@ -0,0 +1,390 @@
...
+}
+
+template <class Parameters>
+void TpAccumulator<Parameters, linalg::GPU>::resetAccumulation(const uint dca_loop) {
+  static util::OncePerLoopFlag flag;
+
+  util::callOncePerLoop(flag, dca_loop, [&]() {
+    resetG4();
+    initializeG0();
+    synchronize();
```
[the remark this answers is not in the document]
gbalduzz: Moved the call_once_err_loop.hpp file and renamed synchronize method in  efaa6ca433e7d3c5c75e001ca8c4a198a51f6e51.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -0,0 +1,434 @@
...
+
+      const CudaComplex<Real> Ga_1 = cond_conj(G_up[i_a + ldgu * j_a], conj_a);
+      const CudaComplex<Real> Gb_1 = cond_conj(G_up[i_b + ldgu * j_b], conj_b);
+
+      const CudaComplex<Real> Ga_2 = cond_conj(G_down[i_a + ldgd * j_a], conj_a);
+      const CudaComplex<Real> Gb_2 = cond_conj(G_down[i_b + ldgd * j_b], conj_b);
+
+      contribution = (Ga_1 * Gb_1 + Ga_2 * Gb_2) * (-factor);
+    }
+      {
```
biddisco: Here and again at line 299 - there is a chunk of (pasted) duplicate code. Can you confirm that this is intentional and not a mistaken cut'n'paste error.

gbalduzz:
Regarding the `callOncePerLoop` functional, the behaviour is constant with `std::call_once`, as it also waits for the call to be completed by any of the threads.

The code is correct as two of the four modes needs to compute two terms of  G(b1,b2) * G(b3,b4) with different indices in frequency and momentum. I will try to reduce code duplication without sacrificing performance, but it can go in a separate PR.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -0,0 +1,434 @@
...
```
PDoakORNL: Ok so this is essentially UpdateG4SpinDifference, but the documentation here is insufficient. This does not really follow the form of tp_accumulator.hpp too clearly for someone who doesn't dream in CUDA. Each case needs to be documented at least to the standard in the other accumulator files. Additionally what purpose are the {} serving here?  Do this instead of working on more new things and I think we will be ready to merge.  Removing the code duplication should wait until later PR after this one is merged.

PR #36, opened 2018-09-14
Remove busy walker wait.

gbalduzz:
This is not originally part of 'gpu_trunk', but it fixes an easy performance bug.
The walker kept on polling the queue, locking its mutex, until it found an element in it. This could actually prevent for some time the accumulator thread to obtain a lock on the mutex.  This way the thread sleeps until an accumulator writes to the queue.


uhaehner:
test this please

PR #34, opened 2018-09-13
Threaded MC runs on the pool.

gbalduzz:
~Depends on #31 for the test.~
Use the thread pool during qmc and allow a walker and accumulator to share the same CPU thread.

PR #42, opened 2018-09-18
Walker async improvement

gbalduzz:
Depends on #34.
Slight performance improvement of the CT-AUX walker using async memory transfers.

### DIFF
--- a/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_walker.hpp
+++ b/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_walker.hpp
@@ -936,9 +936,9 @@ void CtauxWalker<device_t, parameters_type, MOMS_type>::read_Gamma_matrices(e_sp
       }
     }
 
-    vertex_indixes.set(vertex_indixes_CPU, thread_id, stream_id);
-    exp_V.set(exp_V_CPU, thread_id, stream_id);
-    exp_delta_V.set(exp_delta_V_CPU, thread_id, stream_id);
+    vertex_indixes.setAsync(vertex_indixes_CPU, thread_id, stream_id);
+    exp_V.setAsync(exp_V_CPU, thread_id, stream_id);
+    exp_delta_V.setAsync(exp_delta_V_CPU, thread_id, stream_id);
   }
 
   switch (e_spin) {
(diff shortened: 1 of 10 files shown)


biddisco:
Rebased onto master to reduce diffs

biddisco:
test this please

include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_walker.hpp
```diff
@@ -441,6 +441,7 @@ void CtauxWalker<device_t, parameters_type, MOMS_type>::initialize() {
 
 template <dca::linalg::DeviceType device_t, class parameters_type, class MOMS_type>
 void CtauxWalker<device_t, parameters_type, MOMS_type>::do_sweep() {
+  profiler_type profiler("do_sweep", "CT-AUX walker", __LINE__, thread_id);
```
biddisco: You added the profiler in do_sweep, but commented it out in many other places. 
Was that intentional, or just a case of forgetting yo remove the commented out ones before committing?

gbalduzz: No, it is intentional. This function is lengthy and I want it profiled.

src/math/nfft/dnfft_1d_kernels.cu
```diff
@@ -97,7 +97,7 @@ void accumulateOnDevice(const double* M, const int ldm, const int sign, ScalarTy
                         const ScalarType* cubic_coeff, const int size, cudaStream_t stream_) {
   const auto& helper = HelperSelector<ScalarType>::value;
   const static int convolution_size = 2 * helper.get_oversampling() + 1;
-  const auto blocks = getBlockSize(size * size * convolution_size, 128);
+  const auto blocks = getBlockSize(size * size * convolution_size, 85);
```
biddisco: Could you add a comment to say where the magic number of 85 comes from

gbalduzz: Sorry, I was playing around with different multiples of conviultion_size (17 * 5 = 85), and forgot it there. There was not any measurable benefit.
Reverted it to 128 in 82b6ec5fe537316af981c1df0d884e514a0d68f2

uhaehner:
retest this please

PR #45, opened 2018-10-01
MPITypeMap for enumerations.

gbalduzz:
Improvement of #27.
I found a way to pass a general enumeration to our MPI broadcast wrapper, so we can store them directly as parameter members.


uhaehner:
retest this please

gbalduzz:
rebased

PR #46, opened 2018-10-01
Fix hdf5 cpp link error, honour <packagename>_ROOT policy for find_package

biddisco:
For reasons unclear to me, my build started failing due to unlinked symbols in hdf5_cpp since we use the hdf5 c++ interface.

Recent hdf5/cmake find_package combinations seem to do a very poor job of setting up symbols and HDF5_LIBRARIES is empty on my system, but HDF5_C_SHARED_LIBRARY and HDF5_CXX_SHARED_LIBRARY vars do hold the correct vars (if you ask for shared libs). Also HDF5_INCLUDE_DIRS is empty, but HDF5_INCLUDE_DIR hold the correct path. I've added these to the cmake to fix my local build. Hopefully this makes no difference on systems like daint where the module is used to setup paths and the cray wrappers link stuff anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,8 @@ project(DCA++ C CXX)
 
 ################################################################################
 # Set any policies that might be needed
+
+# Adds <packagname>_ROOT variables to search path used by find_<packagename>
 if(POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,12 @@ cmake_minimum_required(VERSION 3.3)
 project(DCA++ C CXX)
 
 ################################################################################
+# Set any policies that might be needed
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)
+endif()
+
+################################################################################
 # Disable in-source builds.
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -7,6 +7,12 @@
 cmake_minimum_required(VERSION 3.3)
 project(DCA++ C CXX)
 
+################################################################################
+# Set any policies that might be needed
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)
+endif()
+
```
gbalduzz: I am not very fluent in CMake. What are these 3 lines supposed to do?

uhaehner:
retest this please

PDoakORNL:
If pycicle daint-gcc-5.3.0-Cx-Lb-Cuda-C6-D-Rm-Pn-T is Cuda six it doesn't even support C++11 so C++14 is surely going to cause trouble.
If that is not what this failure is @biddisco fix it.  Nothing will be merged until the checks actually pass.

CMakeLists.txt
PDoakORNL: It tells cmake to pay attention and not ignore -D<PACKAGENAME>_ROOT or PACKAGENAME_ROOT env variables.  I've noticed it is not always obeyed by the find packages, but apparently it works for hdf5.

PDoakORNL:
retest this please

Issue #5, opened 2018-08-17
Coarsegraining gives wrong results on multiband models.


gbalduzz:
The quadrature integration over the tetrahedra gives wrong results with multiple bands. The test case is a comparison between a single site, two band models (labelled A and B), and a two site, single band model.
It holds G(0) = G_AA + G_AB and G(\pi) = G_AA - G_AB, coming from c_k = (c_A + c_B) / sqrt(2) and c_{k+pi} = (c_A - c_B) / sqrt(2).
The master branch gives the wrong result :

![img1](https://user-images.githubusercontent.com/11310640/44261553-07c98f00-a1e6-11e8-8326-e1cf43c81381.png)
![img2](https://user-images.githubusercontent.com/11310640/44261554-07c98f00-a1e6-11e8-97f9-2bb08bb08898.png)

while the simple summation over q points implemented in 'tambranch' gives the correct result.

![img3](https://user-images.githubusercontent.com/11310640/44261572-1617ab00-a1e6-11e8-8f83-21f8d6e35539.png)
![img4](https://user-images.githubusercontent.com/11310640/44261573-1617ab00-a1e6-11e8-8099-b61f6af7b324.png)

gbalduzz:
Should be fixed by #91.

PR #49, opened 2018-10-08
Simple coarsegraining dca plus

gbalduzz:
Rewrital of the coarse-graining algorithm as per discussion. While this is to be merged after gpu_trunk, I open the PR to let the tests run.

Besides the cleanup of the class the algorithmic differences are:

- Each contribution G(k+q, w) is summed to G(k, w) with uniform weight. Using the previous weights `w_q` does not lead to the correct multiband result.
- In the DCA algorithm, there is no interpolation: H0(k) is computed exactly and stored in memory rather than being interpolated at each function call.
- In The DCA+ algorithm, the interpolated Sigma is stored in memory rather than being recomputed at every call.
- The whole algorithm is parallelized.
- A new parameter "do_simple_q_points_summation()" (default : true) select this new implementation. The old implementation can be used setting this parameter to false. 
- The option of using single precision is removed, as this is no longer a computational bottleneck.


include/dca/function/function.hpp
```diff
@@ -185,16 +185,12 @@ class function {
...
-  template <typename new_scalartype>
-  void operator*=(new_scalartype c);
-  template <typename new_scalartype>
-  void operator/=(new_scalartype c);
+
+  void operator=(scalartype c);
+  void operator+=(scalartype c);
+  void operator-=(scalartype c);
+  void operator*=(scalartype c);
+  void operator/=(scalartype c);
```
gbalduzz: Avoid matching the scalar assignment overload to the function assignment of a non-const function. It also removes one copy.

include/dca/linalg/matrixop.hpp
```diff
@@ -245,8 +245,8 @@ void insertRow(Matrix<ScalarType, CPU>& mat, int i) {
 // Out: ipiv, work
 // Preconditions: mat is a square matrix.
 // Postconditions: ipiv and work are resized to the needed dimension.
-template <typename ScalarType, DeviceType device_name>
-void inverse(Matrix<ScalarType, device_name>& mat, Vector<int, CPU>& ipiv,
+template <typename ScalarType, DeviceType device_name, template <typename, DeviceType> class MatrixType>
+void inverse(MatrixType<ScalarType, device_name>& mat, Vector<int, CPU>& ipiv,
```
gbalduzz: Allows to invert a MatrixView

uhaehner:
Instead of a new parameter "do_simple_q_points_summation()" (default : true), I'd rather add a parameter like "do_fancy_coarsegraining()" (default: false).
Somehow, it seems cleaner to me to have optional features like this with specific paramteres that are turned off by default.

uhaehner:
retest this please

gbalduzz:
If you prefer a default to false then I would name it `use-legacy-coarsegraining`. Is that ok?

uhaehner:
It is not legacy code (unless you define legacy code as code without tests).
It is a sophisticated integration method, which does not support multi-band models at the moment.
The parameter name should reflect the algorithm that is used.
I'd use `Gaussian-quadrature`.

gbalduzz:
It is legacy code: it is some non-optimized code, with old style conventions, that does not produce the correct result in all cases. For the moment we are not going to support it as in fix it, but just leave it to reproduce old results.
I would not call it sophisticated either. The operations involved are the same, with the only differences that weights from Gaussian quadrature are used instead of uniform weights, and these weights do not seem to be correct. The other differences are implementation details and are a questionable parallelization scheme and interpolation of the analytically and cheaply known H_0(k). 
Though use gaussian quadrature is more descriptive. I  renamed it in 17d8ef2e137e25a2dd92e83c5fb07bdc0c08aec3 and updated the description.

uhaehner:
I'm pretty sure that using the Gaussian quadrature points without the proper weights, but uniform weights, is incorrect in general.

gbalduzz:
It seems like the gaussian weights provide the correct result on the new implementation. I will have to further test it, so do not merge yet. It still can be reviewed as the change will be one line plus the update of the baseline files for the tests.

gbalduzz:
retest this please

gbalduzz:
retest this please

gbalduzz:
retest this please

gbalduzz:
I removed the additional parameter as there is not really any reason to keep the old code now. The tp_coarsegraiuning still uses the old coarse-graining, and it will need to be tested for multiband problems in a different PR. This branch provides the correct multiband result and uses the Gaussian quadrature weights. This is ready to be merged IMO when the tests pass (which they do on my local system).

uhaehner:
In my opinion, we should figure out why the old coarse-graining didn't work for multi-band models.
If we don't understand the problem now, I'm pretty sure that it'll catch up with us later.

gbalduzz:
Sure, open an issue. I anyway have to have a look at the 2p part. In the meantime, we can merge this and fix #5 for the main_dca application.

gbalduzz:
retest this, please.

gbalduzz:
New results
simple_coarsegraining_plus_k0
simple_coarsegraining_plus_k1

PR #52, opened 2018-10-23
Clean up StdthreadQmciClusterSolver.

gbalduzz:
Updated the coding style of the threaded solver interface. No effective changes.

### DIFF
--- a/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_accumulator.hpp
+++ b/include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_accumulator.hpp
@@ -49,15 +49,15 @@ namespace solver {
 namespace ctaux {
 // dca::phys::solver::ctaux::
 
-template <dca::linalg::DeviceType device_t, class parameters_type, class Data>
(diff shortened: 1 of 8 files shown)


gbalduzz:
retest this, please.

gbalduzz:
In the coding standard we just have that type names must be upper case. We don't really have a convention for template parameters, and we often have to store them in a class. Maybe something like 
```
template <class  Arg_> 
class Foo {
 using Arg = Arg_
};
```
?

PDoakORNL:
This template parameter naming issue is a little more subtle than I thought. We have frequent cases of subtypes accessing types that are there supertypes template parameters. This asks more of the naming than you might think initially and are why we have Parameters, ParametersType, parameters_type etc. in the code. I think I have something sorted out but I want to see it work to clean up this sort of inconsistency across the code before I propose it.

Certainly this PR is not the place to solve it.

uhaehner:
For me template parameters are just types. So the naming convention for types should apply.
But I'm not against bringing in some consistency when a template parameter is exposed again via a typedef.

PR #53, opened 2018-10-23
Initial sigma

gbalduzz:
- Read only the necessary input to initialize the DCA loop. (Avoid allocating G4 if it is not required)
- Avoid modifying the chemical potential if "adjust-chemical-potential" is not selected.

Tested by first modifying the system-level DCA test to initialize Sigma, and then modifying the DcaDAta member function. 

This will need to be rebased if #49 is merged first.


gbalduzz:
rebased on master

gbalduzz:
retest this please.

PR #61, opened 2018-11-14
Split number of measurements among walkers.

gbalduzz:
Changes the work distribution in the PR solver and gives the option (on by default) to solve https://github.com/CompFUSE/DCA-develop/issues/185. 
Previously the measurements per process were split evenly at the beginning among accumulators.
Now it depends on the parameter `fix_meas_per_walker()`:
- if false the total number of measurements per process is reached independently on the contribution of each walker. Should provide better load balance.
- if true each walker contributes a fixed amount of measurements. It is more statistically safe.


gbalduzz:
I changed the default, added a comment, and updated the complete input in 804d7e0ba724daea5ba7b551689209efa7708c02.
I believe the default value should be true though, but we can discuss about this later.
I am going to update the wiki after the PR is approved.

PR #64, opened 2018-11-22
Papi profiler rewrital.

gbalduzz:
The Papi profiler crashed when used on multiple threads. I ended up rewriting the papi event class and testing it. I would like some feedback from @biddisco if the solution to include event and test only if cmake finds the library acceptable.


uhaehner:
retest this please

PR #77, opened 2018-12-28
Fix #76

### LINKED ISSUE #76: G4 value bug.
For larger than 2x2 clusters the values of G(k1,k2,w1,w2) with negative w1 are not computed correctly, leading to wrong results for G4.

gbalduzz:
Fixes the two particle accumulation on larger than 2x2 clusters. (Issue #76)


include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh
```diff
@@ -9,14 +9,18 @@
 //
 // Helper class for adding and subtracting momentum and frequency on the device.
 
+// TODO: split G4Helper and G4HelperManager into two separate file.
```
gbalduzz: This is done in gpu_trunk2.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -294,47 +298,54 @@ __global__ void updateG4Kernel(CudaComplex<Real>* __restrict__ G4,
...
-	const CudaComplex<Real> Gb_1 = cond_conj(G_up[i_b + ldgu * j_b], conj_b);
-
-	const CudaComplex<Real> Ga_2 = cond_conj(G_down[i_a + ldgd * j_a], conj_a);
-	const CudaComplex<Real> Gb_2 = cond_conj(G_down[i_b + ldgd * j_b], conj_b);
-
-	contribution = -(Ga_1 * Gb_1 + Ga_2 * Gb_2);
-      }
+    case PARTICLE_HOLE_CHARGE: {
+      // contribution <- -\sum_s G(k1, k2, s) * G(k2 + k_ex, k1 + k_ex, s)
+      int w1_a(w1), w2_a(w2);
```
PDoakORNL: As per the style guide I'm not a fan of multiple declarations like this.

test/unit/phys/dca_step/cluster_solver/shared_tools/accumulation/single_sector_accumulation_test.hpp
```diff
@@ -152,20 +152,18 @@ void SingleSectorAccumulationTest<n_bands, n_sites, n_frqs>::prepareConfiguratio
     Configuration& config, Matrix& M, const int nb, const int nr, const double beta, const int n) {
   config.resize(n);
   M.resize(n);
-  dca::math::random::StdRandomWrapper<std::ranlux48_base> rng(0, 1, 0);
+  static dca::math::random::StdRandomWrapper<std::ranlux48_base> rng(0, 1, 0);
```
PDoakORNL: Why make this static?

uhaehner:
Just an idea: Why not test the 2D FT against two 1D FTs?
I understand that the second coordinate transforms as e^{-ikr} but that's just the FT from momentum to real space plus the prefactor 1/N.

test/unit/phys/dca_step/cluster_solver/shared_tools/accumulation/single_sector_accumulation_test.hpp
gbalduzz: To use the same random numbers as the test of the old code, as it used only one function call.

gbalduzz:
@PDoakORNL: Yes, testing the 3x3 case would require another baseline file. There should be no feature that we are missing out from a smaller cluster.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
gbalduzz: Placed on two lines in 0eb8af2b142c8be56475f13b9ccb545da01d7e96.

gbalduzz:
retest this please.

PR #80, opened 2019-01-11
Find fftw library on Summit.

gbalduzz:
Use the correct environmental variable to find FFTW_DIR on Summit and attempt to locate the library if FFTW_LIBRARY is not specified.


uhaehner:
There is another PR in the pipeline that will cover this.

PR #83, opened 2019-01-15
Minor g4 fixes

gbalduzz:
- Fixes comments in tp_accumulator.hpp
- A promised to Ed, removes 'static' from some variables. Running clang-format originated more lines of difference than intended. Lines tp_accumulator_kernels.cu:249-234 should be identical.


PDoakORNL:
So looking at the conflicts in tp_accumulator_kernels.cu. I'm not sure we want the conflicts in this PR.  I believe the conflicted portions were recently changed to fix #76

PR #91, opened 2019-02-09
Phase factors.

### LINKED ISSUE #87: Symmetrization bug.
The current symmetrization code assumes G(b1,b2,r w) = conj[G(b2,b1,r,-w)], while the correct symmetry should be  G(b1,b2,r w) = conj[G(b2,b1,-r,-w)].

gbalduzz:
This PR makes the following changes in this order:
-  Extends the storage of G and G0 in k space to complex numbers.
-  Fixes the Hamiltonian of the two bands chain to be complex. The coarse-graining test fails for systems larger than two sites.
- Fixes the coarse-graining by introducing phase factors in the Fourier transform.  c(k, b) = \sum_r Exp[i k (r + displacement(b) )] c(r, b).
- Fixes #87.


gbalduzz:
retest this please.

gbalduzz:
retest this please.

gbalduzz:
retest this, please.

applications/cluster_solver_check/cluster_solver_check.cpp
```diff
@@ -94,21 +94,26 @@ int main(int argc, char** argv) {
 
     // ED solver
     EdSolver ed_solver(parameters, dca_data_imag, dca_data_real);
-    ed_solver.initialize(0);
-    ed_solver.execute();
-    ed_solver.finalize(dca_loop_data);
+    if (!skip_ed) {
```
PDoakORNL: So what's the purpose of skipping this?  Is this just a debug feature or important for the summit work?  Documentation is necessary.

include/dca/phys/dca_algorithms/compute_free_greens_function.hpp
```diff
@@ -74,18 +76,24 @@ void compute_G0_k_t(
...
-    for (int k = 0; k < KDmn::dmn_size(); ++k) {
-      for (int n = 0; n < OrbitalSpinDmn::dmn_size(); ++n) {
-        for (int m = 0; m < OrbitalSpinDmn::dmn_size(); ++m) {
-          if (g(m, n, k).imag() > 1.e-6)
-            throw std::logic_error("G_0(\vec{k}, \tau) is real!");
-          G0_k_t(m, n, k, t) = g(m, n, k).real();
-        }
-      }
-    }
+    std::copy_n(g.values(), g.size(), &G0_k_t(0, 0, 0, t));
```
PDoakORNL: Taking the address operator on G0_k_t fills me with suspicion, I think this make assumptions about the internal ordering of both functions which is quite suspect.  Refactoring in func::function could easily break this, operator() is returning a reference to a scalar_type no guarantee

test/integration/coarsegraining/coarsegraining_test.cpp
```diff
@@ -39,6 +39,9 @@
 // comparing the bilayer model with the baseline.
 #undef UPDATE_BASELINE
```
PDoakORNL: So this makes it seem like the singleband model is just sitting around untested in the code.  Single_band chain should be removed or both should be tested. It cannot be pushed with changed code and no test.

include/dca/phys/dca_algorithms/compute_free_greens_function.hpp
PDoakORNL: Also there needs to be a copy_n test for func::function that will break as soon as this assumption is violated.

applications/cluster_solver_check/cluster_solver_check.cpp
gbalduzz: I used it during debugging of the new phase factor feature, not really useful for me anymore. I could either remove it, or document the usage in the header.

include/dca/phys/dca_algorithms/compute_free_greens_function.hpp
gbalduzz: The code relies on functions and matrices having column major order. For example include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_cluster_solver.hpp:566. I will add a test to fix the ordering.

test/integration/coarsegraining/coarsegraining_test.cpp
gbalduzz: Ideally the test should run the two singleband and twoband models and compare the results. Unfortunately, there is no way to run two different models in the same execution without heavy refactoring of the cluster initializer. So I run the code with the flag 'UPDATE_BASELINE' producing an hdf5 file with the single-band result, and the test later runs without said flag comparing the twoband results with the baseline. I tried launching a separate process from the test to run the singleband, but I could not get it to work across systems. Do you have any idea on the best way to implement this sort of test?

applications/cluster_solver_check/cluster_solver_check.cpp
PDoakORNL: Choose one, I'd prefer removal

include/dca/phys/dca_algorithms/compute_free_greens_function.hpp
PDoakORNL: At some point it would be good if we did the copy using our own operator constructor or whatever so internal layout wasn't exposed like this.  Not for this PR.

test/integration/coarsegraining/coarsegraining_test.cpp
PDoakORNL: Thinking about your discussion with Urs its perhaps better to document this state of affairs an not introduce some epic hack around it.

applications/cluster_solver_check/cluster_solver_check.cpp
gbalduzz: Removed in 60b9a4137cffd30f3cecf38e02c84ffd93dff07c

include/dca/phys/dca_algorithms/compute_free_greens_function.hpp
gbalduzz: Tested  in 9f7aa28c694fb19450f647fe4dc1df02010bf5e2.
I guess it would be good, but then this is very general and works for any container without having to specialize for Vector, Matrix, some other sort of Matrix, and std::vector. I am not sure we want to remove raw pointers from modern high-performance code, but we can discuss about this later.

test/integration/coarsegraining/coarsegraining_test.cpp
gbalduzz: will do

gbalduzz: added description in a6bc040f3f854eae9be6e1982255c77a5d7e1bdc

PR #95, opened 2019-02-22
Change mpi flags

gbalduzz:
Fixes unit tests on summit (not the integration ones) and allows to use the test scripts to compile with Release flags.


uhaehner:
Setting CMAKE_CXX_FLAGS in an initial cache file doesn't do anything.
Also, gcc on Summit doesn't know about -march=power9, so I removed it completely.
We should ask the CS people (@arghyac007 @yingwaili ), whether we should you use any flags like march, mtune or mcpu.

elwasif:
> Setting CMAKE_CXX_FLAGS in an initial cache file doesn't do anything.
> Also, gcc on Summit doesn't recognize -march=power9, so I removed it completely.
> We should ask the CS people (@arghyac007 @yingwaili ), whether we should use any flags like march, mtune or mcpu.

-mcpu=power9 -mtune=power9 , though the actual performance impact may not be obvious

test/summit_run_tests.lsf
```diff
@@ -36,15 +34,15 @@ source build-aux/summit_load_modules.sh
...
-type=${DCA_BUILD_TYPE:-Debug}
-echo "DCA build type: ${type}"
-cmake -C ../build-aux/summit.cmake -DCMAKE_BUILD_TYPE=$type -DDCA_WITH_TESTS_FAST=ON -DDCA_WITH_TESTS_EXTENSIVE=ON ../
+# Run CMake to configure the build.
+build_type=${DCA_BUILD_TYPE:-Debug}
+printf "Build type: ${build_type}\n\n"
+cmake -C ../build-aux/summit.cmake -DCMAKE_BUILD_TYPE=$build_type -DDCA_WITH_TESTS_FAST=ON -DDCA_WITH_TESTS_EXTENSIVE=ON ../
 
 # Build the tests.
 make -j
```
gbalduzz: I think output on failure is a much more useful option to understand what happened without having to manually rerun the tests.

gbalduzz:
> Setting CMAKE_CXX_FLAGS in an initial cache file doesn't do anything
Would it work to set a variable like `DCA_CXX_FLAGS` and pass it to `CMAKE_CXX_FLAGS` in a CMakeLists.txt file?

test/summit_run_tests.lsf
uhaehner: The detailed output is always written to the test log, which is located in build/Testing/Temporary and for Jenkins runs copied to our project dir.
I don't think it's useful to clutter the test summary with test outputs. It will be quite a mess if you have a test failing that checks a function with thousands of elements.

uhaehner:
> > Setting CMAKE_CXX_FLAGS in an initial cache file doesn't do anything
> 
> Would it work to set a variable like `DCA_CXX_FLAGS` and pass it to `CMAKE_CXX_FLAGS` in a CMakeLists.txt file?

Sorry, I was not really clear on that. The problem was that `CMAKE_CXX_FLAGS` was not a cache variable. That's why the value assigned to it didn't persist.
Anyhow, one should be careful with setting/modifying `CMAKE_CXX_FLAGS`. I think the preferred way is to use `add_compile_options` as it's done in https://github.com/CompFUSE/DCA/blob/master/cmake/dca_compiler_options.cmake.
If we wanted to add `-mcpu=power9 -mtune=power9`, I'd suggest to add something like this to summit.cmake:
`set(ARCH_FLAGS "-mcpu=power9 -mtune=power9" CACHE STRING "TBA.")`
and in dca_compiler_options.cmake:
`if (ARCH_FLAGS)`
`add_compile_options(${ARCH_FLAGS})`
`endif()`

test/summit_run_tests.lsf
gbalduzz: I see, that is quite useful to know, thanks! But the log file still contains all the test output, and I believe it is quite easier to obtain the important information with output-on-failure, especially as it still produces a compact summary at the end of the log, easily accessible with the tail command. Also having the relevant info in the Jenkins email would save the time to log into the cluster find the correct log, unless there is a problem with having too much text dumped it the email.

gbalduzz:
Can this be merged?
I can test the compiler optimization flags in a second moment.

PR #74, opened 2018-12-19
Tweak accumulator memory.

gbalduzz:
First round of improvements in the accumulator's memory consumption.

* Allow Matrix and Vector to use custom allocators. Wrote each currently used allocator in a separate class.
* Define a new cmake option `DCA_WITH_MEMORY_SAVINGS` to compute sequentially the two spin sectors during the tp accumulation. This uses less memory for the intermediate results.

Depends on #62.

PR #98, opened 2019-02-22
Tweak accum memory2

gbalduzz:
Follow up on #74.
Recycle the same memory allocation for intermediate results in the tp accumulation. It is implemented by using shared pointers to GPU matrices in each object performing a step of the accumulation.


gbalduzz:
retest this, please.

gbalduzz:
retest this, please.

PDoakORNL:
Initially I see no test for reshapable matrix?  Tried dropping it into the matrix tests but you have not implemented much of that api.  Since you are using it as a drop in replacement for matrix one might assume at least that. Some calls such as capacity do not carry the same meaning they do for matrix. This is an important class it needs thorough tests.

gbalduzz:
ReshapableMatrix tested in 4245407cc1adaf1cec18f4d77ea6febcb15c2923.

PR #110, opened 2019-03-22
added  clang-format option.

gbalduzz:
Prevents clang from adding `namespace` in a namespace closing comment.


uhaehner:
Automatic namespace comments is probably a nice feature to use even if it adds "namespace".
I'll update the coding style accordingly.

PR #117, opened 2019-05-10
Develop 207

PDoakORNL:
The first of @uhaehner's G4 API change PR's


PDoakORNL:
These were code reviewed in develop so unless they break CI they don't need to be reviewed again.

PR #125, opened 2019-06-17
Fix 116

### LINKED ISSUE #116: MPI collectives fails for large message size.
DCA does not suporet collectives on functions larger than 2^31-1.
Moreover Summit, possibly due to an MPI bug, does not allow to send messages whose size in byte is larger than sizeof(BaseType) * n_scalars, where sizeof(BaseType)  is for example 4 with std::complex<float>.
Fixed in branch https://github.com/CompFUSE/DCA/tree/fix_116.

gbalduzz:
Fix #116, allowing functions of arbitrary size to be processed on Summit.
Remove `factor` from `MPITypeMap` as it is not necessary at the moment, and it would be better to use aggregate MPI if the need arises.


PDoakORNL:
I think it would probably be best to remove the magic '1' from all those calls then.

PR #136, opened 2019-07-22
typical build for cades condo

PDoakORNL:
by request build-aux files for CADES CONDO

### DIFF
--- a/build-aux/cades-condo.cmake
+++ b/build-aux/cades-condo.cmake
@@ -4,8 +4,9 @@
 
 add_compile_options(-fPIC -march=native -mtune=native)
 
-set(FFTW_INCLUDE_DIR $ENV{FFTW_DIR}/include CACHE PATH "FFTW include path")
-set(FFTW_LIBRARY $ENV{FFTW_DIR}/lib/libfftw3.a CACHE FILEPATH "FFTW libary")
+set(FFTW_DIR /software/user_tools/centos-7.2.1511/cades-cnms/spack/opt/spack/linux-centos7-x86_64/clang-7.9.0/fftw-3.3.8-3mx5kwmdcznqlizyos5uipzhpj6jou2w)
+set(FFTW_INCLUDE_DIR ${FFTW_DIR}/include CACHE PATH "FFTW include path")
+set(FFTW_LIBRARY "${FFTW_DIR}/lib/libfftw3.a;${FFTW_DIR}/lib/libfftw3f.a" CACHE FILEPATH "FFTW libary")
 
 set(TEST_RUNNER "mpirun" CACHE STRING "Command for executing (MPI) programs.")
 set(MPIEXEC_NUMPROC_FLAG "-np" CACHE STRING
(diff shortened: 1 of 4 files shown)


build-aux/cades-condo.cmake
```diff
@@ -0,0 +1,29 @@
+# Initial cache list for CADES-CONDO
+#
+# Usage: cmake -C /path/to/this/file /path/to/DCA/source -D<option>=<value> ...
+
+set(CMAKE_CXX_FLAGS "-fPIC -std=c++14" CACHE STRING "compiler flags")
```
uhaehner: I'd prefer using add_compile_options rather than setting CMAKE_CXX_FLAGS directly.
There are often some issues with overwriting with the latter.
Also, C++14 is already set as default.

PDoakORNL: I’d like to migrate to that as well I’ll do that

build-aux/cades-condo.cmake
```diff
@@ -0,0 +1,32 @@
...
+
+# For the GPU support we also need MAGMA.
+#set(MAGMA_DIR $ENV{MAGMA_DIR} CACHE PATH)
+#  "Path to the MAGMA installation directory. Hint for CMake to find MAGMA.")
+
+set(CUDA_TOOLKIT_ROOT_DIR $ENV{CUDA_DIR})
+
+option(DCA_WITH_TESTS_FAST "Enable fast tests" ON)
+
+# Followed by for "llvm7.9.0" i.e. dev llvm with its version set back for nvcc
```
gbalduzz, line 31: do we need this machine generated part?

build-aux/cades-condo_cpu_only_modules.sh
```diff
@@ -0,0 +1,22 @@
...
+spack load gcc/egooyqw
+module load gcc_6/openmpi/3.1.4
+spack load ninja/gzwd46m
+spack load emacs/qp7x25b
+spack load hdf5/4gmsnjn
+spack load cmake/q76ndqk
+spack load git/kibjjo6
+spack load fftw/kpdartc
+export FFTW_DIR=/software/user_tools/centos-7.2.1511/cades-cnms/spack/opt/spack/linux-centos7-x86_64/gcc-6.5.0/fftw-3.3.8-kpdartcqxfk2kdsbcfdtwin75s24z5uu
+export CXX=mpic++
```
gbalduzz, line 21: Is this necessary? On every other system findMPI should be able to link the mpi library without having to resort to wrappers, which  might to be troublesome if we add hpx wrappers.

biddisco: nb.

- hpx does not require wrappers, it uses standard cmake find_package syntax
- findmpi on many systems actually uses the compiler wrapper to do its thing, though these days you can set MPI_CXX_COMPILER to mpic++ and get the same result most of the time. Since that export is only for a particular machine, it's harmless.

PDoakORNL:
this is now obsolete

Issue #124, opened 2019-06-14
Wiki out of date with current input parameters


PDoakORNL:
- [ ] update sp|tp-host
- [x] update four-point channel options

uhaehner:
Related to https://github.com/CompFUSE/DCA-develop/issues/225.

uhaehner:
sp/tp-host changes have not yet arrived in master.
Closing this for now.

PR #137, opened 2019-08-21
Replace/remove deprecated four-point parameter 'type'

uhaehner:
Addresses https://github.com/CompFUSE/DCA-develop/issues/225 and #124.

The wiki has been updated.

@PDoakORNL Do we want to create a minor release for this?

### DIFF
--- a/test/integration/cluster_solver/stdthread_qmci/gpu/stdthread_ctaux_gpu_tp_test_input.json
+++ b/test/integration/cluster_solver/stdthread_qmci/gpu/stdthread_ctaux_gpu_tp_test_input.json
@@ -34,7 +34,7 @@
   },
 
   "four-point": {
-    "type": "PARTICLE_PARTICLE_UP_DOWN",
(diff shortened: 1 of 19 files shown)


gbalduzz:
If we want to maintain backward compatibility of input files, would it not be better to remove four_point_type_ from the parameters, try to read "four-point-type" from the input, and set the appropriate channel flag to true?

uhaehner:
Yes, the intention was backward compatibility and I agree that four-point-type should disappear entirely in the code.
However, I've just figured that this would lead to rather invasive changes from tp-coarsegraining to the ED solver, to the BSE solver.
Since I cannot implement all of these changes now, I propose that we roll back to the old interface while keeping multiple G4 channels under the hood in DcaData and TpAccumulator.

PDoakORNL:
@uhaehner I have no objections to this.

uhaehner:
The wiki page is reverted and updated with the two new channels.
Closing this for now.

Issue #139, opened 2019-08-30
Thread affinity code does not compile on MacOS


ubulling:
The thread affinity code, i.e.
https://github.com/CompFUSE/DCA/blob/master/src/parallel/stdthread/thread_pool/affinity.cpp
 uses linux specific methods and structures such as `sched_getaffinity()` and `cpu_set_t`, which are not available on MacOS.
This might be a possible workaround for MacOS: https://yyshen.github.io/2015/01/18/binding_threads_to_cores_osx.html

uhaehner:
New proposed workaround: no-op for any non-linux system.

ubulling:
Could someone please fix this?

ubulling:
Addressed by PR #185.

PR #140, opened 2019-08-30
Thread affinity workaround for MacOS

uhaehner:
This is a possible fix for #139.
AffinityTest.All still fails on MacOS since I didn't find a way to bind a thread to more the one core.

### DIFF
--- a/include/dca/parallel/no_concurrency/serial_collective_sum.hpp
+++ b/include/dca/parallel/no_concurrency/serial_collective_sum.hpp
@@ -78,9 +78,15 @@ public:
                                   func::function<Scalar, Domain>& f_stddev) const {
     f_stddev = Scalar(0);
   }
+
+  template <typename ScalarOrComplex, typename Scalar, class Domain, class CovDomain>
+  void computeCovarianceAndAvg(func::function<ScalarOrComplex, Domain>& /*f*/,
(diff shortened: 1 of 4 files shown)


uhaehner:
retest this please

PDoakORNL:
What are you using for a build environment on OSX?

uhaehner:
I tested GCC 7.4.0 and Clang 7.0.1.

gbalduzz:
Did @PDoakORNL or @biddisco check if this solves the build on macs? If there is any problem I guess that just doing a no-op for the set affinity is enough. After all, we don't care about small improvement in the scheduling outside Summit or other production machines.

PDoakORNL:
It did not for me. Urs how did you build the compilers and mpi you are using?  Any special cmake flags?

biddisco:
Just FYI. I have not been using a Mac for several years now. My understanding is that the mac does a reasonable job of managing thread resources so that for a non-critical cache intensive task, being switched from one thread to another should not cause too many problems.

uhaehner:
I used the standard macports builds. I didn't use MPI, though. There were some other issues....
I think the best solution as @gbalduzz mentioned is to do a no-op for any non-linux system.
@gbalduzz Could you please fix this?
Closing this now.

PR #157, opened 2019-10-28
DMatrixBuilder and G0 interpolation

gbalduzz:
This PR includes the methods for computing the G0 and D matrix on CPU and GPU, plus all the changes necessary to accommodate them in the codebase.
All the solver .cu files are compiled in the same library to avoid issues at linkage time with nvcc.


gbalduzz:
retest this, please.

PR #158, opened 2019-10-31
Write temporary file between DCA iterations.

gbalduzz:
Write the minimal set of information necessary to restart the DCA loop from the last iteration, in case the time runs out.  
It does not support automatic restart (i.e., you have to manually change the initial self-energy file and number of iterations).


gbalduzz:
closed via #178

PR #159, opened 2019-11-18
Single precision ctaux

gbalduzz:
Subset of the tensorcore branch including only the option to run the CT-AUX solver in single precision.

We need to agree on an acceptable solution for:
 - Where to inject the option: currently it is a template argument of Walker and Accumulator, while the Solver class reads the type from a configuration file, to not change the signature of the solver in main_dca.cpp file. Instead, it might make sense to have a new template argument for the solver, defaulted to double.
- How to test it? Is it ok to have unit tests in both single and double precision and integration tests running only in the selected precision? Do we want two Jenkins configurations?


gbalduzz:
retest this, please.

uhaehner:
retest this please

uhaehner:
retest this please

PR #160, opened 2019-11-25
Optimize ctaux

gbalduzz:
~~Depends on #159~~

As discussed in the conf call, comparing CT-INT and CT-AUX I found some inefficiencies in both.
Regarding CT-AUX

- Some memory was transferred from non-pinned host memory, meaning implicit synchronization would take place.
- Several O(expansion_order * submatrix_size) operations on the CPU where avoidable: as the submatrix size could very well be O(sqrt(expansion_order)) and the submatrix update itself is O(expansion_order**3), these operations are not negligible. [](url)
  - Instead of deleting the configuration vector from the middle, I swap it with the last element, update the pointers and pop_back()
  - New non-interacting vertices are created only when needed.
  - Rows and columns of N are copied instead of swapped, when the last rows/columns are removed.

I attach a profile before and after the changes.
[timeline_st_.zip](https://github.com/CompFUSE/DCA/files/3888889/timeline_st_.zip)

Unfortunately changing the order of the configuration means that I had to update the baseline of all the no-change tests. To check the correctness of the patch, you can run the statistical tests.


cmake/dca_compiler_options.cmake
```diff
@@ -21,7 +21,7 @@
 set(DCA_WARNINGS -Wall -Wextra -Wpedantic -Wno-sign-compare -Wno-dangling-else)
 
 # Languange standard
-set(DCA_STD_FLAG -std=c++14)
+set(DCA_STD_FLAG -std=c++17)
```
gbalduzz: Forgot this one. In the end I only need to not make a separate .cpp file for a single static variable of  CudaProfiler. I can make it, but we might also consider the upgrade given that all the dependencies on Summit are met.

uhaehner:
retest this please

uhaehner:
retest this please

PDoakORNL (approved):
This PR does much valuable cleanup and simplification in addition to optimization.

PR #162, opened 2019-11-27
Optimize ctint configuration.

gbalduzz:
Apply optimizations similar to #160 to the ct-int configuration.

- Much saner interface to the configuration, designed just for the submatrix update.
- O(1) retrieval of the matrix indices and vertex removal.
- Avoid repeated allocation.

The access to a random partner in case of multiple updates is still not optimized (see comment in code), but it should happen rarely and only in certain applications and would require either a fancy data structure, or relying on the statistical uniformity of std::hash.


cmake/dca_compiler_options.cmake
```diff
@@ -21,7 +21,7 @@
 set(DCA_WARNINGS -Wall -Wextra -Wpedantic -Wno-sign-compare -Wno-dangling-else)
 
 # Languange standard
-set(DCA_STD_FLAG -std=c++14)
+set(DCA_STD_FLAG -std=c++17)
```
gbalduzz: as in #160: not strictly necessary, but it would be nice.

uhaehner:
retest this please

cmake/dca_compiler_options.cmake
PDoakORNL: I think while you are here is a good time to make this transition.

include/dca/phys/dca_step/cluster_solver/ctint/structs/ct_int_matrix_configuration.hpp
```diff
@@ -28,22 +28,44 @@ namespace ctint {
...
     return aux_spin == b.aux_spin && interaction_id == b.interaction_id && tau == b.tau;
   }
 };
 
+struct ConfigRef {
+  ConfigRef() = default;
+  ConfigRef(unsigned _config_id, std::uint8_t _leg_id) : config_id(_config_id), leg_id(_leg_id) {}
+
+  unsigned config_id;   // Index of the interaction in the SolverConfiguration.
+  std::uint8_t leg_id;  // Stores if this is the first or second leg of the interaction.
```
PDoakORNL: So are the only valid values 1 and 2 or is there more to it?

include/dca/phys/dca_step/cluster_solver/ctint/structs/solver_configuration.hpp
```diff
@@ -46,54 +49,29 @@ class SolverConfiguration : public MatrixConfiguration {
   template <class RngType>
   void insertRandom(RngType& rng);
 
-  // TODO: pass output as argument.
+  // Returns at most two candidates. -1 stands for a missing candidate.
```
PDoakORNL: well now it always returns an array of size 2.  So its not going to be -1, update the documentation please.
Documenta removal_rand.

src/phys/dca_step/cluster_solver/ctint/structs/ctint_matrix_configuration.cpp
```diff
@@ -0,0 +1,71 @@
...
+namespace dca {
+namespace phys {
+namespace solver {
+namespace ctint {
+// dca::phys::solver::ctint::
+
+MatrixConfiguration::MatrixConfiguration(const InteractionVertices* H_int, const int bands)
+    : H_int_(H_int), n_bands_(bands), sectors_{Sector(), Sector()} {}
+
+MatrixConfiguration& MatrixConfiguration::operator=(const MatrixConfiguration& rhs) {
```
PDoakORNL: These don't follow the semantics I'd expect for an assignment operator. i.e. H_int_ and n_bands_ are not assigned. Either add those or just use a named function like updateSectors.

gbalduzz: updated in 727f88b2 with an assertion on the const member.

include/dca/phys/dca_step/cluster_solver/ctint/structs/solver_configuration.hpp
gbalduzz: Updated the comment. I want something that returns up to two indices, with no allocation, so I return -1 for a missing index.

include/dca/phys/dca_step/cluster_solver/ctint/structs/ct_int_matrix_configuration.hpp
gbalduzz: 0 and 1. Is the comment confusing? I updated it in 727f88b2

include/dca/phys/dca_step/cluster_solver/ctint/structs/solver_configuration.hpp
PDoakORNL: Ah I see for either one.

PR #163, opened 2019-12-06
Adding distributed_tp_accumulator_gpu_test

weilewei:
This PR adds a missing integration test that combines MPI concurrency and tp_accumulator_gpu test. 
- This test compares the results between locally generated G4 and mpi_allreduce produced G4 and assures the difference is zero. It is also based on the assumption that the seed from all ranks maintains the same.  
- The number of ranks could be changed in its CMakeLists.txt


uhaehner:
test this please

weilewei:
Thanks!

PR #165, opened 2019-12-18
Optimize sp accumulator

gbalduzz:
~~It depends on #159.~~

Stores the helper class in constant memory in a similar fashion to the two-particle accumulator.

Brings the GPU runtime in the performance test from 7.63 ms to 6.02 ms in double precision,
and 3.96 ms in single precision. It also slightly improves the readability.

I also tried updating all values in the convolution windows with a single thread per matrix element, but it results in a pessimization on Summit.

EDIT:
I further optimize it by skipping the update of the furthest convolution point. We can justify it by saying that if we truncate a  kernel at distance 8, and we start from a point at 0.1, instead of accumulating at  -8,-7,...,0,...+8, we accumulate at -7,...,+8, i.e. truncating at the points further from the point than the radius. This brings the size of the convolution to 16, which is half a warp, and runs better on the GPU:
4.96 ms in double and 3.73 ms in single. No numerical difference is observed in the tests.


gbalduzz:
retest this, please

gbalduzz:
retest this, please.

PDoakORNL:
@gbalduzz  From the chain of dependencies this looks like it needs to be next.  I won't have time to look at the conflict until tomorrow.  If they are trivial and you can please resolve them. Thanks

PDoakORNL:
This looks good to me but i would like to verify it on Summit as well before we merge.

PDoakORNL (approved):
Passes on summit fast and extensive on summity, LGTM

PR #169, opened 2020-02-07
Generic container for AoS of multiple types

gbalduzz:
~Depends on #168~
I was playing around with variadic templates and I wrote what I think it is a reasonable way to represent  AoS of different types stored in the same buffer. We can use it to limit the number of CPU-GPU memory transfers when uploading the MC configuration. 
There are probably other places that could benefit, but for the moment I cleaned up a bit the CT-AUX g0 interpolation (all files are self-contained includes) to show how it can be used.


gbalduzz:
I played around with c++17 fold expressions. Compilation time is, unfortunately, the same, or insignificantly better. But it should be much easier to read.

gbalduzz:
retest this, please

gbalduzz:
retest this, please.

include/dca/phys/dca_step/cluster_solver/ctint/walker/tools/kernels_interface.hpp
```diff
@@ -31,6 +31,9 @@ void buildG0Matrix(linalg::MatrixView<Real, linalg::GPU> G0, const int n_init,
                    const bool right_section, DeviceConfiguration config,
                    DeviceInterpolationData<Real> g0_interp, cudaStream_t stream);
 
+template <typename Real>
+Real interpolateSlow(Real tau, int linindex, const DeviceInterpolationData<Real>& g0);
```
PDoakORNL, line 35: Some documentation of why the name is Slow would be good here.

PDoakORNL (changes requested):
To capture a little knowledge please add a sentence or two where indicated.  Overall I think this is cleaner and more clear to me that what replaces.

include/dca/phys/dca_step/cluster_solver/ctint/walker/tools/kernels_interface.hpp
gbalduzz: done in 6c2ca2d784af39324b7098c8297178860697f355

PDoakORNL: Thanks

PDoakORNL (approved):

PR #172, opened 2020-02-25
Parallel ed and non-density interactions

gbalduzz:
Rough parallelization of the G computation loop in the ED solver. Should allow a 2x2 bilayer model to run on summit in the time limit.


gbalduzz:
retest this, please

gbalduzz:
Added  the non-density interaction to the ed solver. Bringing it in line with the ct_int branch.

PDoakORNL:
out of date

PR #177, opened 2020-03-07
HDF5Writer cleanup.

gbalduzz:
- Cleanup of the writer class, relying only on the C++ API.
- Keep datasets and group open, allowing to rewrite datasets and reaccessing the same group
- Cleanup writing of vectors and matrices. (easier to read from hdfview).

TODO:
- ~cleanup reader class as well.~

PR #178, opened 2020-03-07
Write checkpoints

gbalduzz:
~~Depends on #177~~  
Replaces #158

Store the self-energy at every dca iteration, so that the output file can be used to restart the simulation in case of a crash.


gbalduzz:
retest this, please.

gbalduzz:
@PDoakORNL merged master into feature branch.

gbalduzz:
retest this, please

misc/wiki/Parameters.md
```diff
@@ -56,7 +57,7 @@ Write out the &chi;<sub>0</sub> function of the BSE lattice solver.
     {
         "output": {
             "directory": "./T=0.5",
-            "output-format": "HDF5",
```
PDoakORNL, line 59: output-format is just about to have another choice ADIOS2 added, this can't be eliminated.  I don't recall any discussion of removing the JSON capability either.

src/io/hdf5/hdf5_writer.cpp
```diff
@@ -19,16 +19,16 @@ namespace dca {
...
+  std::ifstream ifile(filename);
+  return bool(ifile);
+}
+
 HDF5Writer::~HDF5Writer() {
   if (file_)
     close_file();
 }
 
-bool HDF5Writer::fexists(const char* filename) {
```
PDoakORNL, line 27: This should be kept, checking for an ADIOS2 "file" should be through the specialized interface still.

include/dca/phys/dca_loop/dca_loop.hpp
```diff
@@ -15,13 +15,13 @@
 
 #include <cmath>
 #include <iostream>
+#include <filesystem>
```
PDoakORNL, line 18: We still need to support gcc 7.3.0 and filesystem is only partially available and for that compiler it needs
```c++
#include <experimental/filesystem>
```

test/unit/phys/parameters/output_parameters/output_parameters_test.cpp
```diff
@@ -19,7 +19,7 @@ TEST(OutputParametersTest, DefaultValues) {
   dca::phys::params::OutputParameters pars;
 
   EXPECT_EQ("./", pars.get_directory());
-  EXPECT_EQ("HDF5", pars.get_output_format());
```
PDoakORNL, line 22: see above, output format is still should be selectable.

misc/wiki/Parameters.md
gbalduzz: I see. We did not discuss JSON, but we wanted to add writing the computation state  after every DCA loop. 
It is not possible to overwrite a field on the (rather crude) JSON writer. As none uses it and it is anyway not suited to write an output as large as a DCA run, I thought the easiest thing was to remove it. Avoiding the logging of the state if the writer is json is another easy solution, maybe slightly hackish. Less easy would be using a proper writer from a library.
But the important thing is that it is compatible with the ADIOS2 integration.

src/io/hdf5/hdf5_writer.cpp
gbalduzz: file exist is not a property of an hdf5 object, as it just checks if a file exists or not it should need no specialization for a different writer. I admit that the location of the free function is awkward tough, we should just use `std::filesystem::exists`, do you agree?

include/dca/phys/dca_loop/dca_loop.hpp
gbalduzz: why do we need to support gcc 7? I thought we agreed DCA++ is a c++17 code.

gbalduzz:
@PDoakORNL how do you handle the different writer types with ADIOS? I would like to keep persistent file objects, so ideally I would use virtual functions to read/write. Unfortunately these functions have template arguments, so this is impossible.
I was thinking of having a `Writer` class holding a `std::variant<HDF5Writer, JSONWriter, (ADIOS or whatever)Writer>` and dispatch the calls to `execute`, `open/close_group` with `std::visit`. Is this acceptable?

test/unit/phys/parameters/output_parameters/output_parameters_test.cpp
gbalduzz: reverted in e23561ac5229177383403d00e2abdde3bdca7cc0

gbalduzz:
retest this, please

include/dca/phys/dca_loop/dca_loop.hpp
PDoakORNL: I agree but we're a bit stuck on this one. I was not aware that std::filesystem was an exception for gcc7 until now.

tools/complete_input.json
```diff
@@ -1,7 +1,7 @@
 {
     "output": {
         "directory": "./",
-        "output-format": "HDF5",
```
PDoakORNL, line 4: This needs to be added back, complete_input.json is pretty useful.

PDoakORNL (changes requested):

tools/complete_input.json
gbalduzz: thanks for spotting it. Re-added

PDoakORNL (dismissed):

PDoakORNL (approved):
looks good now. I don't know how portable the signal handler will prove to be but lets be optimistic for now.

PDoakORNL:
test this please

PR #185, opened 2020-04-01
Fix #141

### LINKED ISSUE #141: Affinity Test fails with openmpi 3.1.4 "default" build on Centos 7 x86_64
following on from #139 

on other linux it compiles but for openmpi 3.1.4, Centos 7 x86_64
Works if invoked by
```shell
mpirun -np 1 -map-by node:PE=36 test/unit/parallel/stdthread/thread_pool/affinity_test
```

if invoked by ctest which is equivalent to
```shell
mpirun -np 1 test/unit/parallel/stdthread/thread_pool/affinity_test

...
[ RUN      ] AffinityTest.Count
../test/unit/parallel/stdthread/thread_pool/affinity_test.cpp:50: Failure
      Expected: std::thread::hardware_concurrency()
      Which is: 36
To be equal to: dca::parallel::get_core_count()
      Which is: 1
[  FAILED  ] AffinityTest.Count (0 ms)
...
```

Considering whether to approach this as a dependency issue or an issue with the test.

It's certainly worth trying to encourage users to get their affinity issues fixed but it is a portability and maintenance issue.

**For the default mpich 3.3.1 Centos 7 power9 the test works fine.**

gbalduzz:
The set affinity operations is now a no-op on a non-linux system. The test is pretty minimalistic, but this feature does not affect results, and the selected affinity is just what happens to work best on Summit.

@ubulling, can you test this on a mac and merge it?

### DIFF
--- a/include/dca/parallel/stdthread/thread_pool/affinity.hpp
+++ b/include/dca/parallel/stdthread/thread_pool/affinity.hpp
@@ -19,11 +19,15 @@ namespace parallel {
 // dca::parallel::
 
 // Returns a list of cores id for which the calling thread has affinity.
+// If the macro __linux__ is not defined, returns an empty vector.
 std::vector<int> get_affinity();
 
+// Sets the affinity list of the current thread.
+// If the macro __linux__ is not defined, performs a no-op.
 void set_affinity(const std::vector<int>& cores);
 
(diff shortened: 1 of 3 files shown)


ubulling:
Closes #139.

PR #189, opened 2020-04-05
Ct int solver

gbalduzz:
With this PR, the CT-INT solver will finally be usable.
This PR includes: 
- The solver itself.
- several statistic and no-change tests. This is the bulk of the line count.
- changes to the const-ness and minor interface changes to other solvers' walker and accumulator.
- added template parameter for the accumulator scalar type
- measurements and sweeps-per-measurement can be initialized with a vector to have a different number at each iteration. Input files with a scalar parameter are still supported.


gbalduzz:
retest this please.

gbalduzz:
retest this, please.

gbalduzz:
retest this, please

gbalduzz:
fixed weight initialization in  2d6f92c3bcd05ca3d2665745c8389418b3f566ee

gbalduzz:
Test pass, only failure is the distributed G4 test.

include/dca/linalg/util/cuda_stream.hpp
```diff
@@ -31,10 +31,15 @@ class CudaStream {
   }
 
   CudaStream(const CudaStream& other) = delete;
+  CudaStream& operator=(const CudaStream& other) = delete;
 
   CudaStream(CudaStream&& other) {
     std::swap(stream_, other.stream_);
   }
+  CudaStream& operator=(CudaStream&& other) {
+    std::swap(stream_, other.stream_);
```
PDoakORNL, line 40: this is not what I expect operator= to do.  Also I think std::swap only works with cudastream_t because it is just a typedef of a simple type. I'd like to see this be a friend function swap(CudaStream& a, CudaStream& b)

include/dca/linalg/util/magma_queue.hpp
```diff
@@ -13,39 +13,66 @@
...
+
+#include "dca/linalg/util/cuda_stream.hpp"
 
 namespace dca {
 namespace linalg {
 namespace util {
 // dca::linalg::util::
 
-class MagmaQueue {
+class MagmaQueue : public linalg::util::CudaStream {
```
PDoakORNL, line 28: I think a MagmaQueue has a CudaStream I don't think it is a CudaStream.

include/dca/linalg/util/magma_queue.hpp
```diff
@@ -13,39 +13,66 @@
...
+  }
+
+  MagmaQueue(const MagmaQueue& rhs) = delete;
+  MagmaQueue& operator=(const MagmaQueue& rhs) = delete;
+
+  MagmaQueue(MagmaQueue&& rhs) : CudaStream(std::move(rhs)) {
+    swapMembers(rhs);
+  }
+
+  MagmaQueue& operator=(MagmaQueue&& rhs) {
```
PDoakORNL, line 45: Again the semantics of operator= are pretty well established to not swap.  I'd expect the operator=(Whatever&& rhs) to leave an invalid moved from type.
You're going to need to justify this confusing operator= call to the base class.  This doesn't support using operator= in CudaStream since you aren't getting an operator overload anyway.

Also I see std::move(rhs) I would guess rhs is now invalid but then you call swapMembers.

include/dca/phys/dca_step/cluster_solver/ctint/accumulator/ctint_accumulator.hpp
```diff
@@ -143,26 +145,38 @@ void CtintAccumulator<Parameters, device>::initialize(const int dca_iteration) {
...
+      //  Synchronize sp accumulator streams with walker.
+      event_.block(*sp_accumulator_.get_streams()[s]);
+      //  Synchronize both walker streams with tp accumulator.
+      event_.block(*tp_accumulator_.get_stream());
+    }
+  }
+
   configuration_ = walker.getConfiguration();
-  sign_ = walker.getSign();
+  sign_ = walker.get_sign();
```
PDoakORNL, line 165: Why revert to legacy naming for function?

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator.hpp
```diff
@@ -122,6 +123,10 @@ class TpAccumulator<Parameters, linalg::CPU, DT> {
     return 0;
   }
 
+  linalg::util::CudaStream* get_stream() const {
+    return nullptr;
```
PDoakORNL, line 127: And anywhere we use the result of get_stream we check for nullptr?

include/dca/phys/parameters/mci_parameters.hpp
```diff
@@ -107,6 +110,10 @@ class MciParameters {
     return store_configuration_;
   }
 
+protected:
+  // Resize vector arguments to have the same size as the number of iterations.
+  void inline solveDcaIterationConflict(int iterations);
```
PDoakORNL, line 115: Is this correcting for an input that is conflicting?  If so I'd rather see the application halt informing the user of the conflict.  If yes please write a issue so we go fix I guess mci_parameters outside of this PR.

misc/wiki/Parameters.md
```diff
@@ -409,11 +409,15 @@ Instead of an integer, the string `"random"` can be passed to generate a random
 `"warm-up-sweeps":` integer (20)  
 Number of warm-up sweeps.
 
-`"sweeps-per-measurement":` double (1.)  
-Number of sweeps per measurement.
-
-`"measurements":` integer (100)  
-Number of independent measurements in each Monte Carlo iteration.
+`"sweeps-per-measurement":` vector\<double\> ({1.})
```
PDoakORNL, line 412: see above for my attitude on "correcting" incorrect user input.

include/dca/linalg/util/cuda_stream.hpp
gbalduzz: Move assignment is very often implemented in terms of swap. Using a moved object is UB, so we can leave it in any state. Leaving it in the previous state  of the object it was assigned to, is the easiest way to ensure proper resources deallocation.
cudaStream_t is just a pointer to the actual resources, so std::swap is well defined. I could make an extra swap method either inside the class or outside, but I did not need it.

include/dca/linalg/util/magma_queue.hpp
gbalduzz: We could argue if a MagmaQueue is a CudaStream with extra resources. But according to the Liskov Substitution Principle inheritance is appropriate: everywhere you can use a CudaStream you can also use a MagmaQueue.

include/dca/phys/dca_step/cluster_solver/ctint/accumulator/ctint_accumulator.hpp
gbalduzz: I had to bring the ctint walker interface for the sign in line with the other solvers. Or I could have modified the other two solvers. get/set_<variable name> is accepted in the code style for trivial getter setters, and it seems to be the preferred  way. I admit that my code and  the pre-existing is not very consistent on this and we should agree on a convention to follow for the future

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator.hpp
gbalduzz: No, I was relying on the fact that we never ask to sync or do other operations on a CudaStream with CPU version of the accumulator. I made it safer in 76324771

include/dca/phys/parameters/mci_parameters.hpp
gbalduzz: I want the user to either pass a vector that specifies the measurements for each iteration, or a scalar that is applied for each iteration. I don't really need to fix vectors of the wrong size, even if it is rather natural to cut excessive values or duplicate the last number of measurements (any sane simulation has a non decreasing number of measurements).

include/dca/linalg/util/magma_queue.hpp
gbalduzz: see comment above on move assignment.

include/dca/phys/parameters/mci_parameters.hpp
gbalduzz: Opened issue #214

include/dca/linalg/util/magma_queue.hpp
gbalduzz: I don't see the problem in reusing the move constructor/assignment of the base class and having to define only the behavior of the new members.

include/dca/linalg/util/cuda_stream.hpp
PDoakORNL: My knee jerk is more or less like the below blog post.
http://scottmeyers.blogspot.com/2014/06/the-drawbacks-of-implementing-move.html

However I can see this argument really lost out the common usage in the meantime.

include/dca/linalg/util/magma_queue.hpp
PDoakORNL: I think its pretty clear a CudaStream is a resource a MagmaQueue owns or has a reference to.  When Magma is build on top of AMD's libraries it clearly won't have a CudaStream or be a cudastream.  I don't like to see inheritance used as a way to get simplified access to class members.

efdazedo: AMD HIP tries to follow CUDA. I think there is still "HIP stream"  that works like a cuda stream.

include/dca/phys/dca_step/cluster_solver/ctint/accumulator/ctint_accumulator.hpp
PDoakORNL: This part of our coding style is problematic in that it leaks knowledge of the class implementation, in this case it is a simple getter

include/dca/linalg/util/magma_queue.hpp
PDoakORNL: In other code bases this has led pretty directly to messes like this
class MySolver : public MPIObject, GPUObject, HDF5Object, OtherFunctionsWithState
Where MPIObject contains the communicator pointer and a few trivial member functions that would be more readable as my_mpi.rank() but now appear as rank().

GPUObject contains a stream pointer or cublas hande and some member functions....
Similar access patterns etc.

PDoakORNL: I think it would be more consistent and simple to treat all these resources the same.  After all a MagmaQueue is also a cublas_handle + some stuff or a queue + some stuff.

include/dca/linalg/util/magma_queue.hpp
gbalduzz: you just need to change the implementation of CudaStream (and possibly rename it), this is completely orthogonal from the decision of using composition vs inheritance.
The inheritance has no access to the private members, but it exposes the interface, which I want, and has a well defined downcast that I also need. Of course it could be implemented as composition, but then I need to forward the methods, and use a getter to the stream in the rest of the code. While we don't need runtime polymorphism, I am pretty sure this is a well defined  inheritance, and the added 64 bits of storage for the (unused) vtable pointer is a non issue.

include/dca/linalg/util/cuda_stream.hpp
gbalduzz: Of course it is not as performant as a handwritten function with deallocation and setting to nullptr of the other pointer, but all the performance consideration of the blog are not relevant here, as the issue is "non copiability" rather than performance or memory, and the possibly the only relevant cost is the creating and destroying extra Cuda streams.

gbalduzz:
I still don't understand the crusade against inheritance to represent a simple instance of subtyping,  especially given our PR backlog, but I fixed and merged @PDoakORNL PR removing inheritance.

Issue #184, opened 2020-03-31
wrong configuration sign during initialization.


gbalduzz:
The initial sign of the MC configuration does not account for a factor of \prod_i (-U(i)). This is not relevant when using positive interactions and fixed initial expansion order, but it is possible that this issue causes the sign instability when starting the MC solver with a stored configuration different for each walker.
Also in the CT-AUX solver I can not see where the determinant of the initial M matrix is accounted for. Wouldn't this lead to errors for any system with a sign problem and a non-null initial configuration?

gbalduzz:
Foxed by #190

PR #190, opened 2020-04-06
workaround for #184

gbalduzz:
Temporary patch to at least solve the problem in a default configuration. It is anyway the more sensible default.


PDoakORNL:
test this please

PR #191, opened 2020-04-07
Fix 184

### LINKED ISSUE #184: wrong configuration sign during initialization.

gbalduzz:
Depends on #159 .
Fixes #184 by directly computing the log and the sign of the MC weight (computing directly the weight is likely to produce infinities).


gbalduzz:
Replaced by #194

PR #175, opened 2020-03-02
Autocorrelation

gbalduzz:
Depends on #169 and #159 

- option to compute autocorrelation of the expansion order and G_i,j(r=0, t=0)
- cleaned up the hdf5 writer and keep output file open for logging when needed.
- specialized dca_loop to output to hdf5.

PR #193, opened 2020-04-08
Autocorrelation measurement

gbalduzz:
Depends on ~~#191 #189~~ #178 and #169
Replaces #175 

- Autocorrelation measurement class (a fft version could also be added by storing each measurement, as the per-process storage requirement is not much).
- Walker wrapper to perform the measurements after each sweep.
- Option to dump configuration samples to file.

Measurements supported:
- G(b1, b2, r = 0, t = 0)
- expansion order
- log MC weight.


PDoakORNL:
This had to be hand ported to master.

PR #198, opened 2020-04-22
Enable cmake policy CMP0074 for find package using xxx_ROOT vars

biddisco:
This just makes it easier to get find_package to locate user installed libraries (like FFTW)


biddisco:
I have some more comit for this for FFTW, so don't merge yet

PDoakORNL:
dealt with by other means

PR #201, opened 2020-05-21
Distributed G4 implementation

PDoakORNL:
@weilewei will fill this in


weilewei:
This PR provides an implementation of pipeline ringG algorithm to solve the memory-bound issues in DCA++.
- The ringG algorithm works in this way: for each measurement, each accumulator sends its locally generated G2 to a remote accumulator of same thread_id_ in right-hand side neighbor and receives a corresponding G2 from left-hand side neighbor. After receiving a new G2, this accumulator will update its G4, which is 1/p size of original G4 size, where p is the total number of ranks. The algorithm can only be used when nvlink is supported on hardware resources. 
- Added MPI_Gatherv helper function to accumulate G4 across ranks but only for testing purposes. As distributed G4 function is normally used when G4 cannot fit into one node, thus accumulation won't be possible.
- Added helper functions where needed and they are extensively documented.
- Added wiki and example on distributed-g4-enabled config parameter
- Added a python file for element-wise comparison on g4
- Added a unit test for ringG_tp_accumulator_gpu_test
- Uploaded cvdlauncher.sh file for supporting nvlink option on Sumit

weilewei:
One caveat is that G4 gets full G4 size allocation on CPU at the very beginning [here](https://github.com/CompFUSE/DCA/blob/0947da8829e0f587250dbe9137ce06efbb633770/include/dca/phys/dca_data/dca_data.hpp#L310) through a default constructor [here](https://github.com/CompFUSE/DCA/blob/0947da8829e0f587250dbe9137ce06efbb633770/include/dca/function/function.hpp#L288-L290). However, with distributed G4 version, the full allocation of G4 is unnecessary or even dangerous. Each rank should only allocate 1/p of full G4 size. 

However, the allocation is managed by a generic function, which might be used by other places. I could not simply restrict G4 allocation through the constructor without impacting other functions. Suggestions or PR are welcome. However, this concern shouldn't impact the core part of the distributed G4 algorithm and review. Just saying this out first. Thanks.

weilewei:
> One caveat is that G4 gets full G4 size allocation on CPU at the very beginning [here](https://github.com/CompFUSE/DCA/blob/0947da8829e0f587250dbe9137ce06efbb633770/include/dca/phys/dca_data/dca_data.hpp#L310) through a default constructor [here](https://github.com/CompFUSE/DCA/blob/0947da8829e0f587250dbe9137ce06efbb633770/include/dca/function/function.hpp#L288-L290). However, with distributed G4 version, the full allocation of G4 is unnecessary or even dangerous. Each rank should only allocate 1/p of full G4 size.
> 
> However, the allocation is managed by a generic function, which might be used by other places. I could not simply restrict G4 allocation through the constructor without impacting other functions. Suggestions or PR are welcome. However, this concern shouldn't impact the core part of the distributed G4 algorithm and review. Just saying this out first. Thanks.

Thanks @efdazedo for the suggestion via email. This issue now is fixed via commit 337c1f0.

include/dca/function/function.hpp
```diff
@@ -111,6 +116,10 @@ class function {
   std::size_t size() const {
     return Nb_elements;
   }
+
+  void reset_size(std::size_t Nb_elements_new) {
```
PDoakORNL, line 120: why not the more idiomatic resize?

include/dca/function/function.hpp
```diff
@@ -285,9 +294,16 @@ function<scalartype, domain>::function(const std::string& name)
       size_sbdm(dmn.get_leaf_domain_sizes()),
       step_sbdm(dmn.get_leaf_domain_steps()),
       fnc_values(nullptr) {
-  fnc_values = new scalartype[Nb_elements];
-  for (int linind = 0; linind < Nb_elements; ++linind)
-    setToZero(fnc_values[linind]);
+  if(name.substr(0, 2) == "G4" && distributed_g4_enabled)
```
PDoakORNL, line 297: This switch on runtime type "name" is code smell, most of the chances we have of the compiler enforcing correct code when dealing with a distributed function is lost.  We should find a better way to do this. I'd prefer to have a template class specialization for distributed functions, but I think this is a good bootstrapping step.

include/dca/parallel/util/get_workload.hpp
```diff
@@ -42,6 +42,34 @@ int getWorkload(const unsigned int total_work, const unsigned int n_local_worker
...
+  // if not, ranks with (rank_id < nb_more_work_ranks) has to compute 1 more element than other ranks
+  bool balanced = (total_G4_size % static_cast<uint64_t>(mpi_size) == 0);
+  uint64_t local_work = total_G4_size / static_cast<uint64_t>(mpi_size);
+
+  if(balanced) {
+        offset = static_cast<uint64_t>(my_rank)  * local_work;
+        end  = offset + local_work;
+  }
+  else {
+    int nb_more_work_ranks = total_G4_size % static_cast<uint64_t>(mpi_size);
```
PDoakORNL, line 59: our coding standard generally bans prefix for type or unexplained prefixes.  It might just be a low coffee condition but what does nb mean?

include/dca/function/function.hpp
```diff
@@ -47,7 +50,9 @@ class function {
   // Default constructor
   // Constructs the function with the name name.
   // Postcondition: All elements are set to zero.
-  function(const std::string& name = default_name_);
+  // Special case: when distributed_g4_enabled, G4 related variables only gets
+  // allocation of 1/p of original G4 size, where p = #mpiranks
+  function(const std::string& name = default_name_, const bool distributed_g4_enabled = false);
```
PDoakORNL, line 55: This constructor should have a unit test to insure the expected allocation continues to be made in the future. As well as to guard the default argument value.

include/dca/parallel/util/get_workload.hpp
```diff
@@ -42,6 +42,34 @@ int getWorkload(const unsigned int total_work, const unsigned int n_local_worker
   return getWorkload(local_work, n_local_workers, local_id);
 }
 
+inline void getComputeRange(const int& my_rank, const int& mpi_size,
```
PDoakORNL, line 45: unit test

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh
```diff
@@ -122,6 +126,36 @@ inline __device__ unsigned int G4Helper::g4Index(int k1, int k2, int k_ex, int w
          sbdm_steps_[8] * w2 + sbdm_steps_[9] * w_ex;
 }
 
+inline __device__
+void G4Helper::getComputeRange(const int my_rank, const int mpi_size,
```
PDoakORNL, line 130: We need to start having "device" unit tests but this will have to be covered by integration testing until I setup a framework for that.

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator.hpp
```diff
@@ -101,6 +101,9 @@ class TpAccumulator<Parameters, linalg::CPU> {
   // Empty method for compatibility with GPU version.
   void finalize() {}
 
+  // Empty method for compatibility with GPU version.
+  void ringG() {}
```
PDoakORNL, line 105: conceivably this stub could be useful to nodes with limited memory versus G4 as well. Nothing need go here at least until the GPU version works well.

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp
```diff
@@ -450,15 +493,90 @@ void TpAccumulator<Parameters, linalg::GPU>::finalize() {
...
+  }
   // TODO: release memory if needed by the rest of the DCA loop.
   // get_G4().clear();
 
   finalized_ = true;
   initialized_ = false;
 }
 
+template <class Parameters>
+void TpAccumulator<Parameters, linalg::GPU>::ringG(float& flop) {
```
PDoakORNL, line 513: To be reviewed in detail in future.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -17,6 +18,7 @@
 #include <cuda.h>
```
PDoakORNL, line 18: this files is missing #include <mpi.h> which the MPI_WORLD below requires.  In my build environment this breaks the build.

include/dca/function/function.hpp
weilewei: Sure, I renamed it to resize.

include/dca/function/function.hpp
weilewei: Good point! I will make it compile-time and let me research it on how to do it more elegantly.

include/dca/parallel/util/get_workload.hpp
weilewei: Agreed. What I tried to say is total number of ranks that need more work than others. Or maybe rename it to `n_more_work_ranks` or shall I write some comments around this variable?

include/dca/function/function.hpp
weilewei: Not sure if I understand it, but what does it mean "the expected allocation continues to be made in the future."? Can you explain it a bit more? So I can better prepare a unit test for it. Thanks.

include/dca/parallel/util/get_workload.hpp
weilewei: Agreed. I will prepare one soon.

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh
weilewei: Thanks!

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator.hpp
weilewei: Right, @efdazedo also suggested having a cpu version of ringG algorithm. As long as gpu version looks good to the team, I can write one.

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp
weilewei: Sure, I can walk you through the algorithm if needed.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
weilewei: Added, so does mci_parameters.hpp

weilewei:
> Once the include issue in mci_parameters.hpp is fixed I will merge this.
> It's good enough for a first cut and well protected behind its feature switch.
> I will file an issue for the unit tests, which I am serious about and ideally would have been written as soon as the function signature was known.

Thanks for your review! I think I added missing headers now and I will prepare some unit tests mentioned in your comments soon. Let me know if you have any other suggestions. Thanks!

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh
efdazedo: Just a reminder to be careful in using 64bit integer expressions to avoid potential overflow with 32bit integer type. I think  'unsigned int' may be 32bits. Some intermediate arithmetic expressions may also need to be computed in 64bit. Note a complex*16 array with 2^32  entries take up  about 64 GBytes of memory.

weilewei: @efdazedo right, I forgot the device code also needs the overflow protection. I updated: https://github.com/STEllAR-GROUP/DCA/blob/distG4_pr/include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh#L130-L158

PDoakORNL: I agree with @efdazedo but I think we need an issue filed and separate PR's to deal with the integral types used for indexes across a good many locations in the code. For this PR lets leave those you have added uint_64t since that is still an easy type to grep for in the code base.

include/dca/parallel/util/get_workload.hpp
PDoakORNL: I also believe there is an off by one error here. I will add a PR for you to look at @weilewei

include/dca/parallel/util/get_workload.hpp
PDoakORNL: I think more_work_ranks is fine.

include/dca/function/function.hpp
PDoakORNL: And remember we're c++17 now so it may be easier than it would have been.

include/dca/parallel/util/get_workload.hpp
PDoakORNL: PR filed against your distG4_pr on STELLAR-GROPU:distG4_pr

weilewei: Thank you!

include/dca/parallel/util/get_workload.hpp
weilewei: Ok, just fixed. Thanks

include/dca/function/function.hpp
PDoakORNL: I'm going to be working in function soon so I will explain address it then. This is fine for this PR.

include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_cluster_solver.hpp
```diff
@@ -452,11 +452,21 @@ void CtauxClusterSolver<device_t, Parameters, Data>::collect_measurements() {
...
         auto& G4 = data_.get_G4()[channel];
+        // function operator = will reset this G4 size to other G4 size if they are not equal
         G4 = accumulator_.get_sign_times_G4()[channel];
-        if (compute_jack_knife_)
-          concurrency_.leaveOneOutSum(G4);
+        if(parameters_.distributed_g4_enabled())
+        {
+          // do nothing, no accumulation should be performed as G4 size cannot fit into one GPU
+          // reserve this function for testing purpose only
+//            concurrency_.gatherv(G4, concurrency_.first());
```
gbalduzz, line 461: alignment

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/g4_helper.cuh
```diff
@@ -122,6 +126,36 @@ inline __device__ unsigned int G4Helper::g4Index(int k1, int k2, int k_ex, int w
...
+    // if not, ranks with (rank_id < more_work_ranks) has to compute 1 more element than other ranks
+    bool balanced = (total_G4_size % static_cast<uint64_t>(mpi_size) == 0);
+    uint64_t local_work = total_G4_size / static_cast<uint64_t>(mpi_size);
+
+    if(balanced) {
+        offset = static_cast<uint64_t>(my_rank)  * local_work;
+        end  = offset + local_work - 1;
+    }
+    else {
+        int more_work_ranks = total_G4_size % static_cast<uint64_t>(mpi_size);
```
gbalduzz, line 145: same comment as above

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp
```diff
@@ -450,15 +493,90 @@ void TpAccumulator<Parameters, linalg::GPU>::finalize() {
...
+    int my_concurrency_id, mpi_size;
+    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
+    MPI_Comm_rank(MPI_COMM_WORLD, &my_concurrency_id);
+
+    MPI_Request recv_request_1, recv_request_2;
+    MPI_Request send_request_1, send_request_2;
+    MPI_Status status_1, status_2, status_3, status_4;
+
+    // get rank index of left and right neighbor
+    auto mod_op = [](int id, int mpi_size) {return id % mpi_size; };
```
gbalduzz, line 532: what is the need of aliasing the % operator?

include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp
```diff
@@ -450,15 +493,90 @@ void TpAccumulator<Parameters, linalg::GPU>::finalize() {
...
+    // 1) At each time step, local rank receives a new G2 from left hand neighbor,
+    // makes a copy locally and uses this G2 to update G4, and
+    // sends this G2 to right hand neighbor. In total, the algorithm performs (mpi_size - 1) steps.
+    // 2) This algorithm currently requires parameters in input file, please refer to mci_parameters.hpp:
+    //      a) #walker == #accumulator and shared-walk-and-accumulation-thread = true;
+    //      b) and, local measurements are equal, and each accumulator should have same #measurement, i.e.
+    //         measurements % ranks == 0 && local_measurement % threads == 0.
+    for(int icount=0; icount < (mpi_size-1); icount++)
+    {
+        MPI_Irecv(G_[0].ptr(), (G_[0].size().first)*(G_[0].size().second),
```
gbalduzz, line 546: all of this should be hidden in a concurrency subclass. The scalar type of G and G4 can either be complex<double> or complex<float>

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -179,24 +196,63 @@ __global__ void updateG4Kernel(CudaComplex<Real>* __restrict__ G4,
...
+   // originally flattened one dimensional G4 array will be evenly divided by total number of
+   // mpi ranks into different regions (range length of each region is equal up to 1). Since
+   // each rank only allocates its own portion of G4, so offsetting index is needed. Each rank
+   // only computes G4 elements within correct starting and ending index range, otherwise, returns.
+    const uint64_t local_g4_index = static_cast<uint64_t>(blockIdx.x) * static_cast<uint64_t>(blockDim.x)
+                                    + static_cast<uint64_t>(threadIdx.x);
+
+    // get global G4 index
+    uint64_t start, end;
+    g4_helper.getComputeRange(my_rank, mpi_size, total_G4_size, start, end);
```
gbalduzz, line 226: there is no need to recompute start and end at every kernel launch. They can be computed once and passed as kernel parameters, or stored inside the g4_helper.

src/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_kernels.cu
```diff
@@ -179,24 +196,63 @@ __global__ void updateG4Kernel(CudaComplex<Real>* __restrict__ G4,
...
+    id_i = g4_index % size;
+    id_j = (g4_index - id_i)/size % size;
+    id_z = ((g4_index - id_i)/size-id_j)/size;
+
+    // offset global G4 index to local
+    g4_index = local_g4_index;
+  }
+  else
+  {
+    id_i = blockIdx.x * blockDim.x + threadIdx.x;
```
gbalduzz, line 245: will have a second look later, but it seems reasonable. Still, we really need an unit test with a distributed accumulation of G4.

efdazedo:
Perhaps there is an important use case where the distributed G4s can fit on a single node (6 GPUs) but not on a single GPU. It would be nice to have a sub-communicator for all MPI ranks on the same compute node. The within node communication in the ring algorithm would then be performed by memory-to-memory copies without accessing the Network.  Perhaps use MPI_get_processor_name() to find MPI ranks on the same node. Just a thought.

efdazedo:
One idea is to consider  merging the CPU and GPU kernel as one code. Perhaps something like   a collapsed loop   to get "id_i" and "id_j"   using   the following for say the simple  computeGSinglebandKernel()


for( ij=ij_start; ij < (n_rows*n_cols); ij += ij_size) {   
       id_j = ij/n_rows;
       id_i = (ij - id_j*n_rows); 
       assert( ij == (id_i + id_j * n_rows)  

where  for execution on CPU, we have  ij_start == 0, ij_size == 1,  for execution on GPU, we have
ij_start = blockIdx.x * blockDim.x + threadIdx.x; // global thread number
ij_size = blockDim.x * gridDim.x; //   total number of threads

Similar minor code changes for other cuda kernels.  Note the GPU Kernel will work correctly even if the  total number of threads is less than (n_row * n_col).  Just a thought.

include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_cluster_solver.hpp
```diff
@@ -452,11 +452,21 @@ void CtauxClusterSolver<device_t, Parameters, Data>::collect_measurements() {
     if (accumulate_g4) {
       for (int channel = 0; channel < data_.get_G4().size(); ++channel) {
         auto& G4 = data_.get_G4()[channel];
+        // function operator = will reset this G4 size to other G4 size if they are not equal
         G4 = accumulator_.get_sign_times_G4()[channel];
-        if (compute_jack_knife_)
-          concurrency_.leaveOneOutSum(G4);
+        if(parameters_.distributed_g4_enabled())
```
gbalduzz, line 457: This check is missing in line 287

include/dca/function/function.hpp
```diff
@@ -32,6 +32,9 @@
 #include "dca/util/pack_operations.hpp"
 #include "dca/util/type_utils.hpp"
 
+#include "dca/parallel/util/get_workload.hpp"
+#include "mpi.h"
+
```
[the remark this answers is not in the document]
weilewei: When I `#include "dca/parallel/mpi_concurrency/mpi_concurrency.hpp"` in `function.hpp`, I got cyclic dependency error. 

```
Scanning dependencies of target dca_hdf5
[  6%] Building CXX object src/io/hdf5/CMakeFiles/dca_hdf5.dir/hdf5_reader.cpp.o
In file included from /gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_concurrency.hpp:26,
                 from /gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/function/function.hpp:35,
                 from /gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/io/hdf5/hdf5_reader.hpp:23,
                 from /gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/src/io/hdf5/hdf5_reader.cpp:13:
/gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp:53:12: error: 'dca::func::function' has not been declared
   void sum(func::function<scalar_type, domain>& f) const;
            ^~~~
/gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp:53:26: error: expected ',' or '...' before '<' token
   void sum(func::function<scalar_type, domain>& f) const;
                          ^
/gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp:55:24: error: 'function' in namespace 'dca::func' does not name a template type
   void sum(const func::function<scalar_type, domain>& f_in,
                        ^~~~~~~~
/gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp:55:32: error: expected ',' or '...' before '<' token
   void sum(const func::function<scalar_type, domain>& f_in,
                                ^
/gpfs/alpine/proj-shared/cph102/weile/dev/src/distG4/DCA/include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp:55:8: error: 'template<class scalar_type, class domain> void dca::parallel::MPICollectiveSum::sum(int) const' cannot be overloaded with 'template<class scalar_type, class domain> void dca::parallel::MPICollectiveSum::sum(int) const'
   void sum(const func::function<scalar_type, domain>& f_in,

```

include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp
```diff
@@ -294,6 +307,26 @@ void MPICollectiveSum::localSum(func::function<scalar_type, domain>& f, int id)
...
+    int my_rank, mpi_size;
+    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
+    MPI_Comm_rank(MPI_COMM_WORLD, &my_rank);
+
+    if(my_rank == 0)
+    {
+        std::cout << "\n *********performing MPI Gatherv Sum************* \n";
+    }
+
+    my_rank == 0 ? gatherv_helper(f.values(), f.values(), f.get_domain().get_size(), id)
```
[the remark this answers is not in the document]
weilewei: Actually opposite. f.size() is rank local size, and f.get_domain().get_size() is global size of G4.

include/dca/parallel/util/get_workload.hpp
```diff
@@ -42,6 +42,39 @@ int getWorkload(const unsigned int total_work, const unsigned int n_local_worker
   return getWorkload(local_work, n_local_workers, local_id);
 }
 
+/** This returns the first and last linear indexes, not the last + 1
+ *
+ *  i.e. write for(index i = 0; i <= end; ++i) ...
```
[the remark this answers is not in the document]
PDoakORNL: This one is my fault I wrote this and its a different projects convention.

PDoakORNL:
So here's how I'm considering this PR.  It's an initial implementation of the G4 distribution from which to develop the architecture we will use for distribution more generally and develop a good enough and maintainable implementation of the G4 distribution. At this point it needs to be understandable, pass unit tests, not be missing significant unit tests and the "production" code needs to be protected from it by a feature switch.  Since we have another collaboration I'd like to keep in sync with it to some degree I'd like to get this merged as soon as it does no harm to the production code.

I think @gbalduzz 's general comment is a good punch list for future PR's and we need to copy it somewhere, but non of these issues should be addressed in this PR.

PDoakORNL:
I'll have a PR for you soon that fixes a build issue I see with the circular dependence involving mpi_concurrency.

include/dca/function/function.hpp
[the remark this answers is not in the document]
gbalduzz: you don't include mpi in any way, you template on the Concurrency, and the Concurrency provides size and id.

include/dca/function/function.hpp
```diff
@@ -109,7 +114,11 @@ class function {
     return Nb_sbdms;
   }
   std::size_t size() const {
-    return Nb_elements;
+    return nb_elements_;
+  }
+
+  void resize(std::size_t nb_elements_new) {
+    nb_elements_ = nb_elements_new;
```
gbalduzz, line 121: what is the point of this function? this looks fishy.

include/dca/parallel/mpi_concurrency/mpi_collective_sum.hpp
```diff
@@ -166,6 +173,12 @@ class MPICollectiveSum : public virtual MPIProcessorGrouping {
   template <typename T>
   void sum(const T* in, T* out, std::size_t n, int rank_id = -1) const;
 
+  // Gather results accross ranks on process 'rank_id'
```
gbalduzz, line 176: rank_id is called root_id in the parameters

include/dca/function/function.hpp
weilewei: In tp_accumulator_gpu.hpp, we default allocate G4 with full size on CPU from base class https://github.com/CompFUSE/DCA/blob/master/include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp#L164 and copy G4 vector from GPU to CPU here: `https://github.com/CompFUSE/DCA/blob/master/include/dca/phys/dca_step/cluster_solver/shared_tools/accumulation/tp/tp_accumulator_gpu.hpp#L453`. If the size of source and destination is not matched, an error will occur. 

In distributed G4 case, G4 size from both GPU and CPU size are both reduced to 1/ranks. However, the function class does not support templated initialization with distributed G4 parameters. So I added a helper function to function class to resize function size.

weilewei: It wasn't an elegant solution though. However, in DcaData class, G4 has been already allocated and initialized on CPU in here: https://github.com/STEllAR-GROUP/DCA/blob/distG4_pr/include/dca/phys/dca_data/dca_data.hpp#L308-L313. Is it possible to replace BaseClass::G4_ with DcaData::G4_?

gbalduzz:
retest this, please

PDoakORNL:
I've checked compatibility with rb-tree and I think this is safe to go in now.

PR #208, opened 2020-07-29
Distributed G4 fixes

### LINKED ISSUE #207: CPU only build is broken
The merge of #201 the CPU build does not work: The CPU Ctaux solver is defined only for no distribution (due to the two particle accumulator member), while both no distribution and mpi are supposed to be instantiated in the main executable.

gbalduzz:
fixes #207 and the check on the mci parameters.


PDoakORNL:
This is a good improvement and can be considered Approved, but in the interest of sorting out the CI I've moved it to a branch with intervening PR's merged. see #211

PR #209, opened 2020-07-30
Don't run multi GPU test if enough GPU's aren't present

PDoakORNL:
this will prevent multi gpu tests from running if multiple GPU's are not present

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,11 @@ set(DCA_LIBS
   cuda_utils
 )
 
+set(SYSTEM_GPU_COUNT 0)
+
 if (DCA_HAVE_CUDA)
+  EXECUTE_PROCESS(COMMAND bash -c "nvidia-smi -L | awk 'BEGIN { num_gpu=0;} /GPU/ { num_gpu++;} END { printf(\"%d\", num_gpu) }'"
+                  OUTPUT_VARIABLE SYSTEM_GPU_COUNT)
   list(APPEND DCA_LIBS
     blas_kernels
     dnfft_kernels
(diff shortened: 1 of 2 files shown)


efdazedo:
Do you mean the ringG algorithm would fail  if number of GPU < 3? One may think the ringG algorithm with non-blocking MPI isend/irecv would work even if there is only 1 MPI rank.  I think it is also possible for multiple MPI ranks to share the same GPU.

CMakeLists.txt
```diff
@@ -147,7 +147,11 @@ set(DCA_LIBS
   cuda_utils
 )
 
+set(SYSTEM_GPU_COUNT 0)
+
 if (DCA_HAVE_CUDA)
+  EXECUTE_PROCESS(COMMAND bash -c "nvidia-smi -L | awk 'BEGIN { num_gpu=0;} /GPU/ { num_gpu++;} END { printf(\"%d\", num_gpu) }'"
```
gbalduzz, line 153: The ring algorithm should be independendent from the numbner of GPUs on the node. Rather the technical issue is that we rely on cuda aware MPI. Can we have a check similar to hostname==summit for the moment? Assuming automatically detecting if MPI is cuda aware is more complicated

PDoakORNL: The cvd script will definitely have undefined behavior if there is not 1 GPU per rank.

efdazedo: In FAQ for Open-MPI, there is mention of "Can I tell at compile time or runtime whether I have CUDA-aware support?" . It seems to test for MPIX_CUDA_AWARE_SUPPORT
https://www.open-mpi.org/faq/?category=runcuda#mpi-cuda-dev

Just FYI

PDoakORNL: Added #210 and with that I think this should go in so #208 and #206 can pass CI and go in.

PR #219, opened 2020-09-10
Ct int system level test

gbalduzz:
system level test (full DCA loop) for ctint, similar to the equivalent CTAUX test.


weilewei (approved):
LGTM! Thanks!

PR #220, opened 2020-09-11
Fix daint build

gbalduzz:
The Daint build was broken after an update to cuda removed support for the spefic version we were using. I found that it is much easier to rely on system modules where possible rather than maintaining easy-build just for Daint, as it mixes system modules with very few self built tools. Lets see if this works from the jenkins environment.


PDoakORNL (approved):
LGTM @gbalduzz should I worry about the wider tolerance on the analysis test?

gbalduzz:
@PDoakORNL Seems like there is some small difference on the eigenvectors with the new lineal algebra stack. But it should not be a problem, and the c++ analysis code is not really used.  At least for multiband problems.

PR #223, opened 2020-09-24
fix pure virtual function calls involving MPIConcurrency and gtest

PDoakORNL:
Nice example of Kernighan's law here. 

MI isn't to be taken lightly virtual base classes more so. That's all still there and should be stripped away but this takes care of the issue with gtest.


gbalduzz:
Thanks a lot! So it seems that the DcaMpiTestEnvironment does essentially nothing regarding google tests, and a global (unique) pointer to a Concurrency instance would have sufficed, is that right?

gbalduzz (approved):

PR #224, opened 2020-10-01
Properly construct view of constant matrix.

gbalduzz:
I learned the proper way to construct safe views of constant objects: no const_cast shenanigans required.


weilewei (approved):
LGTM!

PR #230, opened 2020-10-26
assert with message if there are no interacting sites

PDoakORNL:
In current master a puzzling segv will occur in debug, and generally in release. This occurs if for install you have
"single-band-hubbard-model" instead of "single-band-Hubbard-model" in your input.

### DIFF
--- a/include/dca/phys/models/general_interaction.hpp
+++ b/include/dca/phys/models/general_interaction.hpp
@@ -76,6 +76,9 @@ void general_interaction<parameters_type>::set_vertex(
 
   // Get a random pair of correlated spin-orbitals.
   const int pos = rng() * correlated_orbitals.size();
+  // This is instead of crashing on segv when module is unknown.
+  // \todo catch earlier in release as well.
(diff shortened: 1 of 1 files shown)


include/dca/phys/models/general_interaction.hpp
```diff
@@ -118,6 +121,7 @@ std::vector<int> general_interaction<parameters_type>::make_correlated_orbitals(
         for (auto b_j : interacting_orbitals) {
           for (int s_i = 0; s_i < SpinDmn::dmn_size(); ++s_i) {
             for (auto b_i : interacting_orbitals) {
+              // This 1.e-3 seems like a very magic number!
               if (std::abs(H_interaction(b_i, s_i, b_j, s_j, delta_r)) > 1.e-3) {
```
gbalduzz, line 125: Indeed, we need to pick only the non-zero elements. But a check against zero (or something like 10^-10) would do, as no arithmetic is performed on H_interaction

gbalduzz (approved):

PR #199, opened 2020-04-24
Print MC weights

gbalduzz:
Depends on #193, #165, #178.
Prints to file the MC weights for the markov chains in the master rank.
Not particularly urgent, but it can be a nice feature for roughly evaluating if the MC thermalized properly even if the autocorrelation time is not computed. 
To be rebased once #193 is merged.

PR #233, opened 2020-12-07
Complex g0

gbalduzz:
Depends on #226, #169, #199

Implements: 
- support for complex Hamiltonians and complex MC.
- support for non spin symmetric models, by storing all the band dependencies into a single electron band sector. Tp accumulation is supported only for particle-particle-up-down.
- implementation of the Rashba model, up to a reorder of bands in the two particle function.

This branch is based on ct-int and has a lot of messy merges. I will probably rebase and squash the history once we have all the dependencies in.


PDoakORNL:
functionality required numerous other PRs to establish as correct and working with real G0 builds.

PR #245, opened 2021-06-23
fix for CADES build and tutorial scripts

PDoakORNL:
Fix bad CUDA architecture for cades.cmake.  Modify tutorial scripts to be more easily run.


PDoakORNL:
current scripts for cades

PR #248, opened 2021-10-29
Initial HIP port

PDoakORNL:
First AMD support PR.  AMD support is still experimental and should be considered broken for the full application.
Some unit tests still fail.

build with DCA_WITH_HIP=1

Should have no effect on DCA_WITH_CUDA=1 builds.


PDoakORNL:
All tests pass on summit.  Mergiing and moving on.

PR #253, opened 2021-11-05
Mpi workaround

frobnitzem:
This patch aliases `MPI_CXX_*` types to their corresponding `MPI_C_*` types to work around non-compliance of MPICH 3.4 with the MPI 3.1 standard (https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node48.htm#Node48).


frobnitzem:
This PR has been updated with a special cmake option to enable the mpich fix:

`-DDCA_FIX_BROKEN_MPICH=ON`

PDoakORNL (approved):
LGTM

PR #274, opened 2022-03-16
Statistical tests fixes

PDoakORNL:
In support of full fix for #271


PDoakORNL:
This was included in other merges for s of q omega demo.

PR #277, opened 2022-07-21
Added model file for Kagome lattice Hubbard model

maierta:
Implemented as a 3-orbital model on a triangular lattice


PDoakORNL:
test this please

PR #322, opened 2024-01-29
Frontier Rocm 5.6 build

PDoakORNL:


PDoakORNL:
Rocm56 was a dead end, same issues as 5.7

PR #327, opened 2024-04-01
fix Rashba Hubbard segfault

PDoakORNL:
Removed strangely justified static const, with llvm this causes a segfault.


PDoakORNL:
This is superseded by more targeted merges.

PR #330, opened 2024-10-16
get adios2 working on frontier post rocm6

PDoakORNL:


PDoakORNL:
ctint tests will be fixed in next PR.

PR #332, opened 2024-11-06
link older libfabric

PDoakORNL:
this could fix the accumulator hang

ccmake is unfortunately no longer usable cause with alt libfabric


PDoakORNL:
ccmake is now usable again it seems a conflict between the hdf5 module and adios2 module is at the heart of this.

PR #337, opened 2025-01-23
use reduce and bcast on root node vs all_reduce

PDoakORNL:
This is an attempt to work around what looks like an MPI hang on frontier.


PDoakORNL:
This just muddies the waters so master will stick with MPIAllReduce for now

PR #359, opened 2026-05-26
Diagonal Disorder

PDoakORNL:
With the approximation of treating the k or r in the disordered system during accumulation as if translational symmetry was intact.


PDoakORNL:
Superseded by #365

Issue #300, opened 2023-03-03
Input Error Messages


PDoakORNL:
Currently DCA does little checking of user input for validity or match with the built executable model etc.  This needs to improve even for expert users.

1. If working directory is not present hdf5 output crashes with vague error.
2. Model sections for models other than the one the executable is built for are silently ignored.
3. Model sections with name typos are silently ignored and model is run with defaults.
4. Numeric inputs are not checkedd to be numeric data resulting in crashes due to random numerical errors later.

tylersax:
@PDoakORNL for problem 2 described above, is "multiple model sections in the input file" something to throw on, or just warn? 

I see test cases in the code that have multiple model sections and would fail if we hard throw any time we read multiple model sections (e.g. `test/unit/pays/dca_step/cluster_solver/shared_tools/accumulation/tp/input_1x1_multi.json `). I see two options (but please add a third if I'm missing something):

1. Having multiple model sections is a legitimate use case (say the user wants to sweep across parameters/models, but otherwise keep the same input file). The program should warn at runtime, but proceed. 
2. Having multiple model sections is a bad pattern. We should throw, forcing the user to narrow down to one. This route will require splitting up test cases that have multiple model sections.

tylersax:
Per offline discussion, we are going to pursue approach 1. That is, we will warn in the case where extra model sections are included but not built against. 

Relatedly, we discussed the issue of input files that have no model parameters at all. We will treat this as similar to bug 3 described above, and will throw in these cases since it is very unlikely that the user intends to run DCA++ with no model parameters.

PR #361, opened 2026-05-28
phys/parameters: throw on missing output directory (issue #300 bug #1)

tylersax:
## Summary
A non-existent output.directory previously surfaced only as a cryptic HDF5 crash. Add OutputParameters::validate(), called from Parameters::readInput on first rank only, which throws std::invalid_argument with a clear message naming the missing directory.

Validation is kept separate from OutputParameters::readWrite so the parsing path stays disk-free and unit-testable; this also avoids needing the existing ReadAll fixture to point at a real path. Follow-up PRs for bugs 2–4 are expected to add validate() to other parameter sections under the same convention.

## Test plan 

- [x] New `InputValidationTest.MissingOutputDirectoryThrows` exercises a JSON fixture pointing at a nonexistent path; expects `std::invalid_argument`.
- [x] All 16 existing parameter unit tests pass locally 
- [x] CI build + test against `origin/master` post-rebase.


test/unit/phys/parameters/input_validation/input_validation_test.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2026 ETH Zurich
```
PDoakORNL: This test needs to be added to the existing output_parameters_test.cpp file.  The convention in DCA++ is one test file per class.  The json input also needs to be moved.

PDoakORNL (changes requested):
Except for the extra test file and directory this looks good to me, once that is fixed this looks to resolve #300 1.

tylersax:
Great - thanks for the pointer. Moved both the test and the .json fixture to their appropriate home.

PDoakORNL (approved):
This looks good now.

tylersax:
CI is passing. Ready to merge when you are, @PDoakORNL

PR #363, opened 2026-06-03
detect model-section input mistakes (issue #300 bugs #2 & #3)

tylersax:
A DCA++ executable is compiled for one model and reads one "*-model" section. Previously a typo'd or missing model section was silently ignored: ModelParameters::readWrite swallows the missing-group exception and the run proceeds on default model parameters.

Detect this at parse time, while the parsed tree is still live:
  * Instrument the JSON reader to track which top-level groups were read. JSONObject gains a mutable accessed_ flag (set on successful getGroup); JSONGroup::childGroupAccess() / JSONReader:: topLevelGroupAccess() report {name, accessed} as ChildGroupStatus. getGroup also switched from operator[] to find(), so a lookup miss no longer inserts a null entry.
  * Add checkModelSections (model_section_check.hpp), a standalone helper invoked from readInput for the JSON path. It reacts to three cases:
      - typo / wrong file (model section present, none read) -> throw std::invalid_argument;
      - multi-model file (built model read, others present) -> warn; - no model section at all -> warn, so model-agnostic uses of Parameters keep working.

The check is compile-time guarded to JSONReader, since the HDF5 reader does not implement a model-section input guard (justified, since HDF5 input is, presumably, machine-generated.

Tests: reader-level access tracking in json_reader_test, and unit coverage of all checkModelSections branches in a new model_section_check_test.


PDoakORNL (dismissed):
This is real improvement of the behavior wrt model sections in the input.

tylersax:
ACK the partial CI failure. I'll dig in see what's going on, then report back here.

PDoakORNL:
It looks like the a couple of gpu tests have outlier input files, you may be able to figure it out without being able to build CUDA or it may be easier to get you onto the section CI/development machine tomorrow.

tylersax:
Based on a quick look, I think this is a simple fix. The input validation is working exactly as intended: it's throwing because the input file for this GPU test _does not contain a model section for the model it's building an executable against_ (and is therefore silently falling back to default params, which is the thing we don't want.)

So I think the clear answer is we need to update those test fixtures to explicitly include the model section, bringing them up to par with the newly introduced input validation scheme. I'll write that up tomorrow and see if it passes. (I'll also try the CUDA build on my GPU box to see if I can repro on my own, that would be handy)

PDoakORNL (approved):
LGTM the tests breaking our rules for model sections have been fixed.  More may appear in the extensive tests but there are other failures to address there so I'd consider it out of scope for this PR.

Issue #364, opened 2026-06-15
Errors during compilation with 'CT-INT' and 'SS-CT-HYB' solvers


cannotcodealot:
Hello, I'm new to DCA++ and trying to build the application using the ```CT-INT``` and ```SS-CT-HYB``` solvers with the following cmake command
```
cmake ../../dca_source \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DDCA_CLUSTER_SOLVER=CT-INT \ # or SS-CT-HYB alternatively
  -DCMAKE_C_COMPILER=mpicc \
  -DCMAKE_CXX_COMPILER=mpicxx \
  -DBLAS_LIBRARIES="-lmkl_intel_lp64 -lmkl_sequential -lmkl_core" \
  -DLAPACK_LIBRARIES="-lmkl_intel_lp64 -lmkl_sequential -lmkl_core" \
  -DTEST_RUNNER=srun \
  -DDCA_WITH_CUDA=OFF
```
and environment

```
Compiler: GCC 11.4.0
MPI: Intel OneAPI MPI 2021.12.0 (via Spack)
OS: Ubuntu 22.04 (zen4)
```

While ```CT-AUX``` compiles successfully, both ```CT-INT``` and ```SS-CT-HYB``` fail at around 96% during the compilation of ```main_dca.cpp```.

In the case of ```CT-INT``` I get the following errors during compilation:
```
...
[ 96%] Linking CXX static library libsignals.a
[ 96%] Built target signals
[ 96%] Building CXX object applications/dca/CMakeFiles/main_dca.dir/main_dca.cpp.o
...
In file included from /home/mkraemer/DCA/dca_source/include/dca/phys/dca_step/cluster_solver/ctint/walker/ctint_walker_choice.hpp:17,
                 from /home/mkraemer/DCA/dca_source/include/dca/phys/dca_step/cluster_solver/ctint/ctint_cluster_solver.hpp:37,
                 from /home/mkraemer/DCA/builds/build_ctint/include/dca/config/dca.hpp:43,
                 from /home/mkraemer/DCA/dca_source/applications/dca/main_dca.cpp:17:
[...]

PR #365, opened 2026-06-16
Implemented disorder in the DCA Loop using the CT-AUX solver

anirudha-mirmira:
## Summary

Support DCA calculations for systems with static on-site disorder. The crux of the implementation is the idea that the addition of disorder configurations breaks the translational symmetry and requires the cluster problem to be solved with two real space indices (for the 2D case) instead of a single displacement index. The disorder averaging, which is carried out analogously to the CPA, restores translational symmetry, thereby allowing for the solution of the cluster problem with a single displacement and subsequently a single momentum index.

The flow proceeds as follows: Subsequent to the calculation of the cluster excluded G0(d), the real space indices are unfolded to get G0(r1,r2), the disorder configurations (box or binary) are added. This disordered_G0(r1,r2) is the bare line the walkers uses to walk the chains, followed by an two index M(r1,r2) accumulation. The resulting interacting G(r1,r2) is disorder averaged, folded back into a single displacement index G(d), transformed to G(k) and fed into the cluster finalization step.

## CMake Options
The new code has an associated `-DDCA_WITH_DISORDER=ON` gate, which defines `DISORDERED_G0` flag. The disorder parts are separated with `#ifdef DISORDERED_G0` conditionals. The `-DDCA_WITH_DISORDER=OFF` flag results is identical to the clean code except in the places as noted below

## Main differences from the DCA++ main branch
These are the changes to existing, shared code paths:

- **`g0_interpolation` now uses the two r-index signature G0(r1, r2) in both the clean and
  disordered cases.** : The clean code calculated a single index as subtract(r2,r1) in the g0_interpolation routines, which is now calculated after inputting both the indices separately;
 
- **Fixed a latent divide-by-sign bug in `local_G_k_w` (used by `computeErrorBars`)** in
  both branches: the `get_accumulated_sign` was used where the scalar accumulated
  `get_accumulated_phase` was intended (missed by the Feb-2023 Phase refactor).

- CMake files where a new  **CMake option `DCA_WITH_DISORDER` (OFF by default)** defining `DISORDERED_G0`, plus a
  new `DisorderParameters` block (with MPI broadcast) and disorder-related parameter wiring.

## Features
- Generates disorder configuration (equally weighted) using random numbers picked from a  **binary** or **box** distribution. Additionally, in the binary case, a `unique-configs: true` parameter allows for the generation of strictly unique configurations
-  The disorder generation works even in the MPI case with a broadcasted seed, which ensures each rank solves the QMC for an identically disordered ensemble. 
-  Error bars and stat tests: The disordered case generates a disorder averaged (translationally invariant) G_k_w and S_k_w per rank allowing calculation of error bars and statistical testing
- Input handling: A missing disorder block in the input is handled elegantly with fallback to the clean limit calculation. Improperly configured `num-configurations: 0` is handled with a warning and calculation with a single configuration

## Example disorder block in input
Disorder is configured by adding a `disorder` block to the run's JSON input. The presence of
the block is what enables the disorder flow; omit it for a clean run.

```json
"disorder": {
  "distribution": "binary",
  "potential": 0.1,
  "density": 0.5,
  "num-configurations": 4,
  "unique-configs": false
}
```

- `distribution` - `"binary"` or `"box"` (any other value is a hard error).
- `potential` - disorder strength: V for binary (levels ±V/2), or the full box width W.
- `density` - fraction of sites at +V/2 for the binary distribution (ignored by box). For values away from 0.5, `adjust-chemical-potential` in the `physics` block should be `false` to prevent chemical potential adjustment from undoing the effect of the shifted average potential.
- `num-configurations` - number of disorder realizations in the ensemble (equal weights).
- `unique-configs` - if `true`, enforce distinct binary realizations (no-op for box).

## Tests
Added and tested disorder tests gated on `DCA_WITH_DISORDER AND NOT DCA_WITH_CUDA AND NOT DCA_WITH_HIP`, which test the following: 
- parameter parsing
- walker-reads-correct-G0
- accumulation and orientation of the r1,r2 unfolding
- full disorder loop
- two statistical tests (U=4 V=0 vs. clean ED reference; box+binary ensemble at V=0.1 - sanity checks).

## Limitations
- **CPU only.** A `DCA_WITH_DISORDER` + CUDA/HIP configuration is a deliberate **hard CMake error**: the GPU version is not implemented. The common G0_interpolation in the GPU case has the original signature and the two index version is not implemented.
- **Single particle quantities only** - G4 averaging and two index handling is not implemented in the disorder path
- **Static, site-local, spin-symmetric disorder only** - The potential enters as a frequency-independent term applied identically to both spins. Bond/off-diagonal, frequency-dependent, and spin-dependent (magnetic) disorder are not supported.
- **Two distributions only** - Only `binary` and `box` are accepted; any other `distribution`
  value is a hard error.
- **Optimization** after profiling, etc. has not been carried out in the case of disorder


anirudha-mirmira:
Commit f5a4f67e69c64c13f89180ba61edb355e1f3a69a also resolves #364

include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_cluster_solver.hpp
```diff
@@ -570,8 +669,13 @@ void CtauxClusterSolver<device_t, Parameters, Data, DIST>::symmetrize_measuremen
   if (concurrency_.id() == concurrency_.first())
     std::cout << "\n\t\t symmetrize measurements has started \t" << dca::util::print_time() << "\n";
 
+#ifdef DISORDERED_G0
+  // No Symmetrize overload for the two-site <nu,R,nu,R,w> domain, and crystal symmetry only holds
```
PDoakORNL, line 673: I believe the symmetrization is also a linear operation and therefore will commute with the sum. One reason not to symmetrize here is the possibility it introduce more floating point error.  The other is we don't have the symmetrization overload written. 

My intuition is that applying the symmetrization to each configuration or once to the summed M will make little difference unless the sampling is insufficient.

include/dca/phys/dca_loop/disorder/makeDisorderConfigurations.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+    const bool is_box = (distribution == "box");
+
+    // Seed from the Monte Carlo seed, shifted by a fixed offset so the disorder realizations
+    // use a stream distinct from the QMC walkers (which derive their seeds from get_seed()).
+    // The seed is read on the root rank and broadcast as part of Parameters, so it is identical
+    // on every rank (every rank generates the same disorder ensemble) and fixed for the whole
+    // run (the ensemble is the same in every DCA iteration, i.e. quenched disorder). No separate
+    // seed broadcast is needed here.
+    constexpr unsigned disorder_seed_offset = 0x9e3779b9u;  // golden ratio; arbitrary but fixed
```
PDoakORNL, line 74: This seems perfectly fine and its useful to know the behavior of Parameters.
but  more import are the assumptions `MakeDisorderConfigurations` makes wrt Parameters, i.e. its requirements of Parameters for this class to function as described.

The following assumptions are made here:
1. The global seed is read from input and an identical value is returned by parameters.get_seed() in all contexts parameters is available.
2. the parmeters.get_seed() value is immutable

One might assume `Parameters` are immutable over a DCA run but the code violates this for instance with chemical potential.

include/dca/phys/dca_step/cluster_solver/ctaux/ctaux_cluster_solver.hpp
```diff
@@ -243,6 +241,13 @@ CtauxClusterSolver<device_t, Parameters, Data, DIST>::CtauxClusterSolver(
 #endif
       averaged_(false),
       writer_(writer) {
+#ifdef DISORDERED_G0
+  // The two-particle (G4) accumulator measures against the clean G0_k_w_cluster_excluded, which is
+  // inconsistent with the disordered walker; fail fast rather than produce wrong G4 under disorder.
+  if (parameters_.isAccumulatingG4())
```
PDoakORNL, line 247: This is much better than having it still go and be wrong.

include/dca/phys/dca_step/cluster_solver/stdthread_qmci/stdthread_qmci_cluster_solver.hpp
```diff
@@ -59,6 +59,7 @@ class StdThreadQmciClusterSolver : public QmciSolver {
   using typename BaseClass::Accumulator;
   using Walker = stdthreadqmci::StdThreadQmciWalker<typename BaseClass::Walker, Data>;
   using SpGreensFunction = typename BaseClass::SpGreensFunction;
+  using SpDisorderedGreensFunction = typename BaseClass::SpDisorderedGreensFunction;
```
PDoakORNL, line 62: This breaks the CTINT build and in a way that is not going to be clear to most people building the code. Some clean way of dealing with this is needed. An intelligible compilation or cmake error would be fine until disorder is supported for ctint.

PDoakORNL (commented):
@anirudha-mirmira  These are my comments based on the work you did which looks good, it would have been nice to retain the ability to run both disorder and standard calculations without recompilation but that can probably be restored.

Most of what I see holding this back from merge is other delta's inherited from the starting branch and its divergence for master. We should discuss directly.

PR #368, opened 2026-06-24
Symmetrization M1: derive orbital-op U_S from H0 (shadow mode)

tylersax:
Derive each point-group operation's orbital transform U_S directly from H0(k), replacing the hand-coded transformationSignOf* lookups, and verify it against the deterministic G0 oracle. All shadow-mode: nothing consumes the record for imposition yet, so production output is unchanged.

`solve_orbital_op_signs.hpp` (new): U_S = D.P, where the band permutation P is the geometry-given symmetry-table image and the +/-1 signs D are solved from the H0 coupling constraints by enumerate-and-verify.

`cluster_symmetry`: add shadow accessors get_mapped_point and get_orbital_op. 

`set_symmetry_matrices`: populate mapped_point (the band-independent momentum image, ".first" promoted out of the symmetry-matrix pair).

Integration test: add `PerOpMapMomentumUS` (ensure U_S is well formed), `GroupShadowCrossCheck` (the H0-verified group equals the declared group on every shipped model), `MappedPointShadow`, and synthetic H0 rejection fixtures covering all three throw branches. Note these are designed under the assumption that we will encounter more complex forms of H0 in the material models. 

Candidates still come from the declared point group, so DCA_point_group is the shadow oracle.

## Mini-roadmap 
Snapshot of where we are in the project (subject to change but this is what I'm thinking):

- ~M0: Characterization harness~ 
- M1: Derive orbital-op U_S from H0 (shadow mode) [👈🏻 You are here!]
- M2: Derive each model's 2D point group from H0 and report divergence from the declaration
- M3: Imposition rewrite — replace the existing transformationSignOfK-based imposition with the derived U_S
- M4: Full-model characterization coverage + material track — honeycomb migration, material models (NiO/FeSn), general complex-U_S for genuine orbital mixing
- M5: Runtime integration + statistical tooling 
- M6: Cleanup

### DIFF
--- a/include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
+++ b/include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
@@ -1,5 +1,4 @@
-// Copyright (C) 2018 ETH Zurich
-// Copyright (C) 2018 UT-Battelle, LLC
+// Copyright (C) 2026 UT-Battelle, LLC
 // All rights reserved.
 //
 // See LICENSE for terms of usage.
@@ -10,17 +9,28 @@
 // Derives the orbital-operation matrix U_S of each point-group operation from H0(k) and populates
 // cluster_symmetry::get_orbital_op(), without reading hand-coded Lattice::transformationSignOf*.
 //
-// Scope: for every analytic model U_S is a real signed permutation U_S = D P, where the permutation
-// P is already encoded in the symmetry table (the band image .second) and the only unknown is the
-// diagonal of +/-1 signs D = diag(sigma). Substituting into the relation H0(k) = U_S H0(k_new) U_S^dagger
-// gives, entry by entry,
+// Scope: for every analytic model, U_S is a real signed permutation U_S = D P, where the
+// permutation P is already encoded in the symmetry table (the band image .second) and the only
+// unknown is the diagonal of +/-1 signs D = diag(sigma).
 //
-//   H0(k)(b0,b1) = sigma(b0) sigma(b1) H0(k_new)(image(b0), image(b1)).
+// The symmetry table stores not the exact momentum image S k but its folded representative
+// k_new = S k - G, and with intra-cell orbital offsets a_b the fold is not free: H0 picks up the
+// diagonal folding phase V(G) = diag(phi_b), phi_b = e^{i G.a_b} (cluster_symmetry::get_fold_phase(),
+// a real +/-1 for every shipped model). Substituting U_S = D P into the symmetry relation
+// H0(S k) = U_S H0(k) U_S^dagger and folding S k back to k_new gives, entry by entry,
 //
-// So each nonzero coupling fixes the product sigma(b0) sigma(b1) to the +/-1 ratio of the two H0
-// entries. The analytic models have at most n_b = 3 orbitals, so once the gauge sigma(0) = +1 is
-// fixed there are at most 2^(n_b-1) <= 4 candidate sign vectors. We simply enumerate them and keep
-// the one that satisfies every coupling constraint. Three conditions make an operation fail:
+//   sigma(b0) sigma(b1) H0(k)(b0,b1) = phi(image(b0)) phi(image(b1)) H0(k_new)(image(b0), image(b1)).
+//
+// So each nonzero coupling fixes the product sigma(b0) sigma(b1) to the +/-1 fold-corrected ratio
+// of the two H0 entries. The phi factors matter: without them the solve absorbs the folding phase
+// into the signs and returns a representative-convention-dependent dressing of U_S (e.g. masking
+// the odd parity of p_x under a mirror at a zone-boundary momentum) instead of the intrinsic
+// orbital transform.
+//
+// The analytic models shipped with DCA++ have at most n_b = 3 orbitals, so once the gauge
+// sigma(0) = +1 is fixed there are at most 2^(n_b-1) <= 4 candidate sign vectors. We simply
+// enumerate them and keep the one that satisfies every coupling constraint. Three conditions make
+// an operation fail:
 //
 //   * a magnitude mismatch (an entry vanishing on only one side) means the permutation is not a
 //     symmetry of H0;
@@ -55,12 +65,15 @@ constexpr double kCouplingZeroTol = 1e-10;
 // magnitude from 1 must both be below this.
 constexpr double kSignRealTol = 1e-9;
 
-// Solves the +/-1 signs of one operation s and writes that op's U_S block into u_s. Throws if op s
-// is not a signed-permutation symmetry of H0. Shared by the whole-group populator -- which lets the
-// throw propagate, since a model must be a genuine symmetry group -- and by the group verification,
+// Solves the +/-1 signs of one operation s and writes that op's U_S block into u_s. fold_phase is
+// the diagonal folding phase phi (callable as fold_phase(k, band, s)) that undresses the folded
+// H0 samples, so the solved signs are the intrinsic orbital transform. Throws if op s is not a
+// signed-permutation symmetry of H0. Shared by the whole-group populator -- which lets the throw
+// propagate, since a model must be a genuine symmetry group -- and by the group verification,
 // which catches it to classify the op.
-template <typename SymFunc, typename H0Function, typename UFunc>
-void solveSignsForOp(int s, int nb, int nk, const SymFunc& sym, const H0Function& H0, UFunc& u_s) {
+template <typename SymFunc, typename FoldFunc, typename H0Function, typename UFunc>
+void solveSignsForOp(int s, int nb, int nk, const SymFunc& sym, const FoldFunc& fold_phase,
+                     const H0Function& H0, UFunc& u_s) {
   // Band permutation P: row b's single entry lands in column image[b]. This is the geometry-derived
   // half of U_S (which orbital maps to which); it is k-independent for a point-group op, so read it
   // at k = 0. Only the signs that decorate it are unknown.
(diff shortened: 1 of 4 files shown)


include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
+// Copyright (C) 2018 ETH Zurich
```
PDoakORNL: If its a new file you only need 2026 UT-Battelle, LLC

include/dca/phys/domains/cluster/cluster_symmetry.hpp
```diff
@@ -54,11 +54,48 @@ class cluster_symmetry<cluster_domain<scalar_type, D, N, R, S>> {
...
+  // uniform symmetry relation reads G(k) = U_S G(mapped_point) U_S^dagger.
+  static func::function<int, mapped_point_dmn_t>& get_mapped_point() {
+    static func::function<int, mapped_point_dmn_t> mapped_point("mapped_point_super_cell");
+    return mapped_point;
+  }
+
+  // Diagonal folding phase V(G) = diag(e^{i G.a_alpha}) per (cluster point, band, super-cell op).
+  // Populated alongside get_symmetry_matrix() in set_symmetry_matrices. Runtime guard rejects (throws)
+  // any phase that is not +/-1, which holds for every shipped (symmorphic, point-symmetry-only) model.
+  // Complex storage to support a genuine non-+/-1 fold is a future generalization.
```
PDoakORNL: This should probably be a \todo so it doesn't get lost.

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
...
+//
+// See LICENSE for terms of usage.
+// See CITATION.md for citation guidelines, if DCA++ is used for scientific publications.
+//
+// Author: Tyler Sax (tylersax@gmail.com)
+//
+// Derives the orbital-operation matrix U_S of each point-group operation from H0(k) and populates
+// cluster_symmetry::get_orbital_op(), without reading hand-coded Lattice::transformationSignOf*.
+//
+// Scope: for every analytic model U_S is a real signed permutation U_S = D P, where the permutation
```
PDoakORNL: I think "model, " There should be a break keep U_S from being seen as the definition of analytical model.

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
...
+// P is already encoded in the symmetry table (the band image .second) and the only unknown is the
+// diagonal of +/-1 signs D = diag(sigma). Substituting into the relation H0(k) = U_S H0(k_new) U_S^dagger
+// gives, entry by entry,
+//
+//   H0(k)(b0,b1) = sigma(b0) sigma(b1) H0(k_new)(image(b0), image(b1)).
+//
+// So each nonzero coupling fixes the product sigma(b0) sigma(b1) to the +/-1 ratio of the two H0
+// entries. The analytic models have at most n_b = 3 orbitals, so once the gauge sigma(0) = +1 is
```
PDoakORNL: Maybe "Our analytical models" I think there is likely some > 3 orbital model that is analytically.

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
...
+  const int n_ops = SymDmn::dmn_size();
+
+  std::vector<int> verified;
+  for (int s = 0; s < n_ops; ++s) {
+    try {
+      detail::solveSignsForOp(s, nb, nk, sym, H0, u_s);
+      verified.push_back(s);
+    }
+    catch (const std::exception&) {
+      // op s is not a signed-permutation symmetry of H0 -> not part of the verified group.
```
PDoakORNL, line 205: If I understand correctly we can only end up here if we're in contradiction with the user input symmetry table then ok I can conceptualization that as an exception path.

test/integration/phys/symmetrization/symmetrize_characterization_test.cpp
```diff
@@ -204,7 +210,43 @@ double maxAbsDiff(const F& a, const F& b) {
...
 }
 
-// TEST 1 of 4: DomainActivation -- sanity gate. Reads the op counts off the symmetry
+// Runs fn and returns the what() string of any std::exception it throws, or "" if it ran
+// to completion. The rejection fixtures use this to assert not just that the sign solver
+// throws, but that it throws for the intended reason (the three branches catch three
+// distinct corruptions and share two exception types, so the message is what distinguishes
+// them).
+template <typename F>
+std::string captureThrowMessage(F&& fn) {
```
PDoakORNL, line 219: Catch2 has a macro construct to deal with this whole situation, and while it does look to me like the ancient version of google test we're using doesn't, are you aware if a more current Gtest does?

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
...
+// magnitude from 1 must both be below this.
+constexpr double kSignRealTol = 1e-9;
+
+// Solves the +/-1 signs of one operation s and writes that op's U_S block into u_s. Throws if op s
+// is not a signed-permutation symmetry of H0. Shared by the whole-group populator -- which lets the
+// throw propagate, since a model must be a genuine symmetry group -- and by the group verification,
+// which catches it to classify the op.
+template <typename SymFunc, typename H0Function, typename UFunc>
+void solveSignsForOp(int s, int nb, int nk, const SymFunc& sym, const H0Function& H0, UFunc& u_s) {
+  // Band permutation P: row b's single entry lands in column image[b]. This is the geometry-derived
```
PDoakORNL, line 77: I think you going to need to take the fold_phase as an argument here too because I think its going to be needed down when the ratio is calculated or U_S will end up "contaminated" by the sign flip from the k-folding with some systems.

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
```diff
@@ -0,0 +1,195 @@
...
+        const bool zero_lhs = std::abs(lhs) < kCouplingZeroTol;
+        const bool zero_rhs = std::abs(rhs) < kCouplingZeroTol;
+        if (zero_lhs && zero_rhs)
+          continue;  // coupling absent on both sides: consistent, but no constraint.
+        if (zero_lhs != zero_rhs)
+          throw std::logic_error(
+              "solveOrbitalOpSignsFromH0: H0 magnitude mismatch for op " + std::to_string(s) +
+              " -- the band permutation is not a symmetry of H0 (a coupling vanishes on only one "
+              "side).");
+        const std::complex<double> ratio = lhs / rhs;
```
PDoakORNL: Here the ratio is going to need something like 
```
const double fold = fold_phase(k, b0, s) * fold_phase(k, b1, s);
const std::complex<double> ratio = lhs / (fold * rhs);
```
At least for the three band example I was looking at.

PDoakORNL (commented):
Ok I didn't have time to completely work through this in full detail but I think there is something up with the folded_phase and solveSignsForOp that is a bit subtle and not correct right now.
But that also depends on whether I understand your intent right from the comments with respect to this method and U_S.  You may have just deferred this to a future PR.

A unit test that GPT-5.5 "extracted" for me from the three band hubbard that hits this 
```
TEST(SolveOrbitalOpSigns, ThreebandBoundaryFoldDoesNotMaskPxParity) {
  struct Sym {
    std::pair<int, int> operator()(int k, int b, int s) const {
      return {0, b};  // one folded k representative: (pi,0), no band permutation.
    }
  };

  struct H0 {
    std::complex<double> operator()(int b0, int, int b1, int, int) const {
      const std::complex<double> I(0., 1.);
      if (b0 == 0 && b1 == 1)
        return 2. * I;  // d-px at k=(pi,0), t_pd=1
      if (b0 == 1 && b1 == 0)
        return -2. * I;  // Hermitian partner as in threeband_hubbard.hpp
      return 0.;
    }
  };

  struct U {
    double data[3][3] = {};
    double& operator()(int r, int c, int) {
      return data[r][c];
    }
  };

  U u;
  dca::phys::detail::solveSignsForOp(0, 3, 1, Sym{}, H0{}, u);

  EXPECT_DOUBLE_EQ(u(0, 0, 0), +1.);
  EXPECT_DOUBLE_EQ(u(1, 1, 0), -1.);  // Fails today: current solver gives +1.
  EXPECT_DOUBLE_EQ(u(2, 2, 0), +1.);
}
```
Well allegedly, I was already going down this path so it may have picked up on this.

I will try and put together a mock setup that actually uses the real model and not llm brute force in the next day or two.

test/integration/phys/symmetrization/symmetrize_characterization_test.cpp
tylersax: Yes, it looks like Gtest offers `EXPECT_THAT(actual_value, matcher)` ([docs](https://google.github.io/googletest/reference/matchers.html)) which would do the job here. Something like `EXPECT_THAT(fn, ThrowsMessage<std::logic_error>(HasSubstr("…")))`. But looks like we're currently several versions behind that.

include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
tylersax: Correct, but note this becomes more load-bearing in the next PR: we will use it to report on a divergence between the declared symmetry and the derived symmetry

tylersax:
@PDoakORNL great catch. I don't fully have my head around it but I do think you're onto something. The fold phases are quite tricky. Let me noodle on this more tomorrow and see if I can get to the bottom of it. 

Regarding all of the other inlines: ack, will address.

tylersax:
OK @PDoakORNL - have a look at this update. It's more clear to me now where the fold phase was sneakily getting absorbed. This commit properly dresses and undresses the sign, and adds a test to pin it down in exactly the way you suggested. So back to you: is there anything else we should be thinking about here? What other geometric edge cases come to mind?

maierta:
I have not looked at the details yet, but Tyler and I discussed the overall concept and came to the conclusion that it makes sense and should work.

PDoakORNL (approved):
Ok this LGTM now.
There is some new issue with the CI which I need to resolve but assuming checks still pass when that is fixed I think this is ready to merge.

tylersax:
CI green, good to merge

PR #369, opened 2026-07-03
Symmetrization M2: derive each 2D model's point group and report divergence from the declaration

tylersax:
We now derive a 2D model's point group: a universal holohedry pool, filtered by the lattice geometry (this PR), then gated on H0 invariance (M1's `verifiedSymmetryOps`). In this milestone the derivation is diagnostic only — production symmetrization continues to use the declared group unchanged, and a report is printed when the two disagree.

**Commit 1** — holohedry pool (holohedries_2d.hpp). Replaces the per-geometry point-group headers (2d_square / 2d_hexagonal / 2d_rectangular / 2d_oblique) with a single header defining the two 2D holohedries (D4 and D6) and the candidate pool they generate. By the crystallographic restriction, every 2D point group is a subgroup of one of these — so the lower-symmetry groups (D2, C2, …) no longer need to be declared per model (the geometric filter derives them). This simpler representation of the candidate groups also addresses a suspected bug in D6: the legacy list omitted the identity (which I believe matters during imposition) and used mirror denominator 6 as opposed to the correct  12 (effect: collapses six mirror lines to three). A completeness oracle (holohedries_2d_test) asserts that filtering the pool by each 2D Bravais lattice reproduces each of the four 2D holohedries (8/12/4/2 ops).

**Commit 2** — derive and report (derive_point_group.hpp). deriveAndComparePointGroup runs a three-stage derivation per cluster family: install the holohedry pool as the live candidate group → the existing geometric search keeps ops preserving the lattice and cluster → an H0-invariance gate (M1's verifiedSymmetryOps) keeps ops admitting a signed orbital permutation U_S with U_S H0(O·k) U_S† = H0(k). It then restores the declared symmetry state — verified, throwing on failure — and returns a report naming any divergence. update_domains calls it after each family's declared initializer and prints on rank 0 only when derived ≠ declared.

**Commit 3** — name the divergent ops. Adds point_group_symmetry_element::describe(), a human-readable label decoded from the operation matrix: Schoenflies name plus geometry, e.g. C4^3 (rotation 270 deg), sigma (mirror @ 45 deg). The divergence report lists the labeled ops in both directions (under-declared and unverified) instead of bare counts, so a report pinpoints exactly which symmetry a model's declaration is missing or overclaiming.

**Scope**: 2D models only (no 3D pool yet); lattices without the initializeH0 interface skip derivation entirely (compile-time detection)

## Mini-roadmap 
Snapshot of where we are in the project (subject to change but this is what I'm thinking):

- ~M0: Characterization harness~ 
- ~M1: Derive orbital-op U_S from H0 (shadow mode)~ 
- M2: Derive each model's 2D point group from H0 and report divergence from the declaration [👈🏻 you are here]
- M3: Imposition rewrite — replace the existing transformationSignOfK-based imposition with the derived U_S
- M4: Full-model characterization coverage + material track — honeycomb migration, material models (NiO/FeSn), general complex-U_S for genuine orbital mixing
- M5: Runtime integration + statistical tooling 
- M6: Cleanup


tylersax:
Converting to draft until M1 #368 is resolved, since it looks like we will have some changes there that need to be incorporated here

tylersax:
@PDoakORNL This PR is ready for review. @maierta and I discussed this morning and are aligned on the approach.

src/phys/domains/quantum/point_group_symmetry_element.cpp
```diff
@@ -80,6 +83,42 @@ void point_group_symmetry_element::transform(const double* t0, double* t1) const
...
+
+  if (DIMENSION == 2) {
+    // O is column major, so its first column (O[0], O[1]) is (cos, sin) of the rotation angle for a
+    // proper rotation (det +1) and (cos, sin) of twice the mirror-line angle for a reflection (det
+    // -1). Each operation gets a Schoenflies name (Cn^k for rotation, and sigma for reflection) and
+    // a geometric description.
+    const double det = O[0] * O[3] - O[2] * O[1];
+    if (det > 0.) {
+      const long raw = std::lround(std::atan2(O[1], O[0]) * 180. / M_PI);
+      const int deg = static_cast<int>((raw % 360 + 360) % 360);
```
PDoakORNL: Not that this doesn't seem to work, but to me it feels strange to be doing this in degrees. I think 12 being the lcm of the crystallographic rotation orders is the right factor to use here. Then I think this code will probably read in a way that aligns better with the group theory for 2 and 3D point groups.

PDoakORNL (changes requested):
Outside of the CI failure which seems to be a simple header file update, the "turn" factor is all I see in this in PR I'd like to see updated.

src/phys/domains/quantum/point_group_symmetry_element.cpp
tylersax: Fair point - calculating in degrees is a little goofy here. Updated to twelfths (*say that word 10 times fast*), but kept the degrees in the print statement since (I think) that's the natural way to read it. LMK if you think we should wipe that too and stick to fractions all the way.

tylersax:
Header issue should be fixed. Good to run CI again.

src/phys/domains/quantum/point_group_symmetry_element.cpp
PDoakORNL: I don't expect to ever see any angle in physics or mathematics expressed in anything but radians but that might just be me @maierta?

tylersax: Rads are perfectly fine by me. Updated. 👍

tylersax:
Updated and ready for CI

PDoakORNL (approved):
Review addressed. LGTM

tylersax:
CI green, ready to hop on the merge train

tylersax:
@PDoakORNL bumping this back up. More code headed your way soon, so it would be good to get these two outstanding PRs merged to keep the commit stack clean.

PR #372, opened 2026-07-16
Add Vpd and Vpp to three-band Hubbard model

PeizhiMai:
Add Vpd and Vpp interaction terms to three-band Hubbard model


PDoakORNL (dismissed):
LGTM

maierta:
Looks good to me too.

PR #373, opened 2026-07-16
Correct finite-Q multiband G4 accumulation

PeizhiMai:
Correct finite-Q multiband G4 accumulation (PARTICLE_PARTICLE_UP_DOWN channel only) for both CPU and GPU code.


PeizhiMai:
> This generally looks good to me, but one question I don't remember the answer to: Did the "plaquette test" pass with these changes?

Yes, the "plaquette test" pass with these changes.

maierta:
Would you be able to write a test that implements the plaquette test?

PeizhiMai:
In my previous plaquette test, I work with converged G and G4. But here you probably want a test using only a few measurements, is that right?

maierta:
This one is good to go as far as I'm concerned

PDoakORNL (approved):
This looks good to me now as well.

PR #374, opened 2026-07-21
Symmetrization: recover symmetries for same-site orbital swap case

tylersax:
**Background**: I was running FeAs against the derived-symmetry report we just landed in #369, and it came back with 4 ops vs the expected 8. The solver was rejecting C4, C4^3 and the two diagonal mirrors.

**Problem**: FeAs's two orbitals (d_xz, d_yz) both sit at a_vec = 0. set_symmetry_matrices builds the band image by transforming r + a_b and looking for an orbital at the image position with the same flavor. That test isn't meaningful for two orbitals on one site -- it matches every band equally. The "flavor" breaks the tie, and that pins each band to itself. The band image comes out the identity for every op. solveSignsForOp took the image on faith and solved only the signs. Signs can flip an entry, but they cannot move it to another band. So the solver falsely rejected real symmetries.

**The (small) fix**: treat the geometric image as a candidate rather than an oracle. solveSignsForOp splits into tryPermutation (same function we had before, returning false where it used to throw) and deriveOrbitalOpForOp, which checks the geometric candidate first and otherwise walks the remaining permutations. Throw only when none of them works.

**Result 1**: FeAs goes from 4 derived ops to 8, and is added as a characterization fixture to pin the number.

**Result 2**: A Kagome win falls out of the same change. Kagome labels its three symmetry-equivalent sublattices with distinct flavors, so its matching finds no admissible image at all and records the -1 that #369 added a guard for. Deriving P from H0 does more than dodge the out-of-bounds read -- it recovers the ops geometry could not place, and Kagome goes from 2 derived ops to 12.

### DIFF
--- a/include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
+++ b/include/dca/phys/dca_step/symmetrization/solve_orbital_op_signs.hpp
@@ -9,9 +9,12 @@
 // Derives the orbital-operation matrix U_S of each point-group operation from H0(k) and populates
 // cluster_symmetry::get_orbital_op(), without reading hand-coded Lattice::transformationSignOf*.
 //
-// Scope: for every analytic model, U_S is a real signed permutation U_S = D P, where the
-// permutation P is already encoded in the symmetry table (the band image .second) and the only
-// unknown is the diagonal of +/-1 signs D = diag(sigma).
+// Scope: for every analytic model, U_S is a real signed permutation U_S = D P. Both halves are
+// derived from H0: the permutation P (which orbital maps to which) and the diagonal of +/-1 signs
+// D = diag(sigma). The symmetry table's band image (.second) is only the first candidate for P;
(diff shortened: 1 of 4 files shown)


PDoakORNL:
I read over your indicated significant commit, looks good to me but lets get this cleaned up so @maierta can also take a look. By which I mean lets get #369 merged so this PR gets reduced to something easy to look at.

PDoakORNL:
@tylersax I don't want to force the update here, can you take a look at doing the update?

tylersax:
Rebased and ready to go 👍

tylersax:
Pulling this back into draft. When I was working on the analysis to calculate symmetry gains, I got a seg fault on Kagome that I didn't look into any further at the time. It does not appear related to any of the symmetry code we added - I suspect it's a long-lived bug that was just suppressed by the fact that kagome has been running with `no_symmetry<2>`. I don't have a full picture yet but since this PR touches kagome, I want to pull it back until I have a clearer sense.

